PR #3, opened 2016-05-26
Add support for multipart uploads

barcharcraz:
I needed multipart upload support for my project so I added it.
Contains four new functions, one for each multipart endpoint and a fourth
that deals with uploading a whole large file

(sidenote: I am soon going to implement an IOStream for an AWS file, I'm not sure
if AWSS3 is a good place for that but I'd be willing to contribute it once I finish, the
idea is that it will send Range headers to seek and read so you can seek around really
big archives)


samoconnor:
Thanks for the contribution @barcharcraz !
An `IOScream` interface for S3 using range sounds like a great addition.
I won't get time to review and merge this for a few days, but your contributions are very welcome.

PR #10, opened 2017-06-20
add support of content-encoding 

jingpengw:
as an option in s3_put


jingpengw:
Just saw that there is already a PR for more general usage.
https://github.com/samoconnor/AWSS3.jl/pull/9

PR #13, opened 2017-09-24
Update REQUIRE

jingpengw:
`Requests` was used in this package.


samoconnor:
Thanks @jingpengw,
At this point the Requests.jl dependancy has been completely removed.

PR #30, opened 2018-07-26
Updates for 0.7, drop 0.6

ararslan:
Testing on AWSCore master.


ararslan:
@samoconnor I think this will need something similar to https://github.com/JuliaCloud/AWSCore.jl/commit/8dbb8287fd2924fd1f1a1ef7eaff31588fbc6ed9, but I'm not sure how exactly to implement that here.

samoconnor:
Looks like here https://github.com/samoconnor/AWSS3.jl/blob/master/src/AWSS3.jl#L558

ararslan:
I guess for consistency this package should also use EzXML anyway.

samoconnor:
And here too: https://github.com/samoconnor/AWSS3.jl/blob/master/src/AWSS3.jl#L676

However, using an XML library to construct the `CompleteMultipartUpload` seems like overkill.

I would prefer to see something simple like this:

```julia
parts = ["""
  <Part>
    <PartNumber>$i</PartNumber>
    <ETag>$etag</ETag>
  </Part>
 """ for (i, etag) in enumerate(parts)]

content = "<CompleteMultipartUpload>$(join(parts))</CompleteMultipartUpload>"
```

samoconnor:
Note; I won't be able to spend much time on this for a few days at least. But I'm happy to merge PRs if you want to continue working on it.

ararslan:
Yes I'll keep at it, thanks!

PR #31, opened 2018-07-31
Replace LightXML with EzXML

ararslan:
Followup from #30. @samoconnor

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,6 +4,6 @@ HTTP
 SymDict
 Retry
 XMLDict
-LightXML 0.7.0
+EzXML
 MbedTLS
 DataStructures
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -29,7 +29,7 @@ using HTTP
 using SymDict
 using Retry
 using XMLDict
-using LightXML
+using EzXML
 using Dates
 
 const SSDict = Dict{String,String}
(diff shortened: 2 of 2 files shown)


ararslan:
Seems that it's the same issue with credentials not being available.

samoconnor:
I'm testing this locally.
There is another issue caused by places in AWSCore that do `String(::Vector{UInt8})` which in 0.7 results in the Vector being truncated to zero length :(

ararslan:
Yeah, because the string takes ownership of the underlying data in the array upon construction. If you can point me to particular instances where it's causing a problem I can submit a PR. For example, if you're getting errors, the backtrace would be helpful.

samoconnor:
https://github.com/JuliaCloud/AWSCore.jl/commit/fda9819a6715e66c97bf1b5c6592a4bd092df981

samoconnor:
I'm now getting as far as `ERROR: LoadError: UndefVarError: BufferStream not defined`

samoconnor:
https://github.com/JuliaCloud/AWSCore.jl/commit/5d173bd159d8e6748c7873ff02fe36cd4c3e1d13

ararslan:
So I take it things are all good now? :D

samoconnor:
The AWSS3 tests passed on my machine with the resent AWSCore commits.

This is looking good too: https://travis-ci.org/samoconnor/AWSS3.jl/jobs/410592595

The `String` thing makes me concerned that there may be other 0.7 change bugs lurking in `AWSCore`. I feel like it needs a code review against a list of things that may break in 0.7.

samoconnor:
Fair enough.
I've just tried out AWSSQS.jl and AWSSNS.jl on my local machine.
Aside from the patch below for the runtest.jl file and a dep warnings, the tests pass.

```patch
diff --git a/test/runtests.jl b/test/runtests.jl
index 31da8b4..3c9d4f2 100755
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -5,6 +5,7 @@
 #==============================================================================#


+using AWSCore
 using AWSSQS
 using AWSSNS
 using Retry
```

I'm happy to accept PRs for AWSSQS.jl and AWSSNS.jl to drop 0.6 and fix the deprecations.
If we can get AWSCore, AWSS3, AWSSQS and AWSSNS all passing CI I'll tag them all.

ararslan:
Sounds fantastic. Thanks!

Issue #12, opened 2017-07-25
s3_exists should use HEAD instead of GETting a small range


s2maki:
GET on a URL causes Requests.jl to parse the resulting data by its content type.  Using GET with a byte range of 0-0 will return a single byte (the first byte in the file), causing such parsing to fail.

Better to use HEAD instead of GET when testing for object existence.

samoconnor:
Thanks @s2maki 

I'm now calling `s3_get_meta` : https://github.com/samoconnor/AWSS3.jl/commit/79b1b47928504a531bcf320fc257d2ae8f5ce706#diff-bfa26c7ba85a0e383f74a605295a9456R210

s2maki:
Thanks!

PR #36, opened 2019-02-13
Use GET request when checking for S3 existence

### LINKED ISSUE #35: s3_get_meta does not get error codes
Since `s3_get_meta` uses a `HEAD` request, it does not get S3 error codes, which are provided in the body of a response. The caller can get back an HTTP status code but has no further information.

The solution is to undo the change that implemented https://github.com/JuliaCloud/AWSS3.jl/issues/12 and go back to `0-0` range requests.

nicoleepp:
Closes #35

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -234,7 +234,9 @@ function s3_exists(aws::AWSConfig, bucket, path; version="")
 
     @repeat 2 try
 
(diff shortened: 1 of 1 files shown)


samoconnor:
Given that #12 was a `Requests.jl` work-around, this seems ok to me.

nicoleepp:
Bump, could this be merged?

PR #41, opened 2019-07-01
added access permissions for s3_put

ExpandingMan:
I have need of the `"x-amz-ac"` header key for `s3_put`.  This field allows one to set permissions of the object to one of [these](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl) pre-configured options.

I have added a unit test for this, but note that I am unable to reproduce the tests locally, so I haven't checked this very well.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -277,15 +277,15 @@ s3_delete(a...; b...) = s3_delete(default_aws_config(), a...; b...)
 - `metadata::Dict=`; optional `x-amz-meta-` headers.
 """
 function s3_copy(aws::AWSConfig, bucket, path;
-                 access_permissions::AbstractString="",
+                 acl::AbstractString="",
                  to_bucket=bucket, to_path=path, metadata::SSDict = SSDict())
 
     headers = SSDict("x-amz-copy-source" => "/$bucket/$path",
                      "x-amz-metadata-directive" => "REPLACE",
                      Pair["x-amz-meta-$k" => v for (k, v) in metadata]...)
 
-    if !isempty(access_permissions)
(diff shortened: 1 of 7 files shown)


src/AWSS3.jl
```diff
@@ -614,13 +619,16 @@ s3_purge_versions(a...) = s3_purge_versions(default_aws_config(), a...)
 # Optional Arguments
 - `data_type=`; `Content-Type` header.
 - `encoding=`; `Content-Encoding` header.
+- `access_permissions=`; 'x-amz-acl' header for setting access permissions with canned config.
```
rofinn: Would you be willing to rename this to just `acl` to more closely match the cli interface?

ExpandingMan: Sure, if you want.  Something I'd like to ask you to consider though: for those who aren't constantly dealing with AWS the name `acl` is extremely mysterious.  I'm not sure I'd remember what that pertains to even justa few days from now.  Are you sure you want to call the keyword arg that?

rofinn: FWIW, ACLs are a pretty common concept outside of S3 as well. Most linux admins would also be familiar with them for managing default access permissions within a directory (which is a bit more intuitive than how S3 does it). The other alternative I can see is just `access_control_list`.

ExpandingMan: Ok, good enough for me then, I'll take your word for it.

I'll change it when I get a chance.

rofinn: Thank you

ExpandingMan:
Sorry that took so long, forgot about it at some point.

rofinn:
Since nobody else has commented on this PR I'm happy to just merge it.

PR #42, opened 2019-07-18
Add an S3Path type

rofinn:
Blocked on tagging FilePathsBase.jl.


Project.toml
```diff
@@ -0,0 +1,28 @@
...
+HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
+MbedTLS = "739be429-bea8-5141-9913-cc70e7f3736d"
+Retry = "20febd7b-183b-5ae2-ac4a-720e7ce64774"
+SymDict = "2da68c74-98d7-5633-99d6-8493888d7b1e"
+UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
+XMLDict = "228000da-037f-5747-90a9-8195ccbf91a5"
+
+[compat]
+AWSCore = ">= 0.5"
+julia = "1"
```
oxinabox: Need compat version for FilePathsBase

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+    )
+end
+
+function S3Path(str::AbstractString; config::AWSConfig=aws_config())
+    str = String(str)
+    root = ""
+    path = ()
+    dir = true
+
+    @assert startswith(str, "s3://")
```
oxinabox: This should be an `ArgumentError`

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+        true,
+        config,
+    )
+end
+
+function S3Path(str::AbstractString; config::AWSConfig=aws_config())
+    str = String(str)
+    root = ""
+    path = ()
+    dir = true
```
oxinabox: Can we call this `isdirectory`?
`dir` to me sounds like it might be the directory part of the name.

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+
+    @assert startswith(str, "s3://")
+    tokenized = split(str, "/")
+    bucket = strip(tokenized[3], '/')
+    drive = "s3://$bucket"
+
+    if length(tokenized) > 3
+        root = "/"
+        # If the last tokenized element is an empty string then we've parsed a directory
+        dir = isempty(last(tokenized))
```
oxinabox: per above
```suggestion
        isdirectory = isempty(last(tokenized))
```

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+        return fp.drive * fp.root
+    elseif attr === :separator
+        return "/"
+    elseif attr === :bucket
+        return split(fp.drive, "//")[2]
+    elseif attr === :key
+        return fp.dir ? join(fp.segments, '/') * "/" : join(fp.segments, '/')
+    else
+        # Call getfield even though we know it'll error
+        # so the message is consistent.
```
oxinabox: Why don't we just delete the top?
```
  if isdefined(fp, attr)
        return getfield(fp, attr)
```

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+        return getfield(fp, attr)
+    end
+end
+
+# We need to special case join and parents so that we propagate
+# directories correctly
+function Base.join(prefix::S3Path, pieces::AbstractString...)
```
oxinabox: I love the fact that we get to do this so much.

the name `joinpath` was a clear indication of `join` begging for the use of dispatch on a path type.

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+    segments = String[prefix.segments...]
+    dir = endswith(last(pieces), "/")
```
oxinabox: `isdirectory`

Is it worth abostracting this?
It is also called in the constryuctior

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+    )
+end
+
+function parents(fp::S3Path)
+    if hasparent(fp)
+        return map(1:length(fp.segments)-1) do i
+            S3Path(fp.segments[1:i], fp.root, fp.drive, true, fp.config)
+        end
+    else
+        error("$fp has no parents")
```
oxinabox: Why doesn't this return itself.
after all `..` in the unix root directory leavies you in in the unix root directory

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+        key = fp.key
+    else
+        return false
+    end
+
+    objects = s3_list_objects(fp.config, fp.bucket, key; max_items=1)
+    return !isempty(objects)
+end
+
+Base.real(fp::S3Path) = fp
```
oxinabox: Can we comment this with reference to the FilePathBase docs?

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+    if exists(fp)
+        !exist_ok && error("$fp already exists.")
+    else
+        if hasparent(fp) && !exists(parent(fp))
+            if recursive
+                mkdir(parent(fp); recursive=recursive, exist_ok=exist_ok)
+            else
+                error(
+                    "The parent of $fp does not exist. " *
+                    "Pass recursive=true to create it."
```
oxinabox:
```suggestion
                    "Pass `recursive=true` to create it."
```

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+                )
+            end
+        end
+
+        write(fp, "")
+    end
```
oxinabox: Should return the newly created directory here.

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+    if isdir(fp)
+        k = fp.key
+        # Only list the files and "dirs" within this S3 "dir"
+        objects = s3_list_objects(fp.config, fp.bucket, k; delimiter="")
+
+        # Only list the basename and not the full key
+        basenames = Set(s3_get_name(k, string(o["Key"])) for o in objects)
+
+        # Lexographically sort the results
+        return sort!(filter!(!isempty, collect(basenames)))
```
oxinabox: Why do we construct a `Set` only to `collect` it?
if we mean `unique` then lets just say that?

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+            error("S3 path $object is not empty. Use `recursive=true` to delete.")
+        end
+    end
+
+    @debug "delete: $fp"
+    s3_delete(fp.config, fp.bucket, fp.key)
+end
+
+function Base.readdir(fp::S3Path)
+    if isdir(fp)
```
oxinabox: Do we need
```suggestion
    if FilePathsBase.isdir(fp)
```
?
or is that the same as `Base.isdir`.
If it is, then I have mild objections to the earlier part where we overloiaded `FilePaths.isdir`
and it actually was changing e `Base.isdir`

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+
+        # Lexographically sort the results
+        return sort!(filter!(!isempty, collect(basenames)))
+    else
+        throw(ArgumentError("\"$fp\" is not a directory"))
+    end
+end
+
+Base.read(fp::S3Path) = s3_get(fp.config, fp.bucket, fp.key)
+Base.read(fp::S3Path, ::Type{String}) = String(read(fp))
```
oxinabox: We can probably get rid of this?
Since the new syntax is to hust call `String(read)`?

Or if we must keep this then we should just define this in `FilePathsBase`

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+function Base.write(fp::S3Path, content::Union{String, Vector{UInt8}})
+    s3_put(fp.config, fp.bucket, fp.key, content)
+end
+
+function FilePathsBase.mktmpdir(parent::S3Path)
+    fp = parent / string(uuid4(), "/")
+    mkdir(fp)
+    return fp
```
oxinabox: Can delete this line if we fix `mkdir` to return the directory it mase

test/runtests.jl
```diff
@@ -258,31 +260,32 @@ versions = s3_list_versions(aws, bucket_name, "key3")
 @test s3_get(aws, bucket_name, "key3") == b"data3.v3"
 
 
-HTTP.ConnectionPool.showpool(stdout)
+# HTTP.ConnectionPool.showpool(stdout)
```
oxinabox: Why this all commented out?

test/s3path.jl
```diff
@@ -0,0 +1,115 @@
...
+        test_ismount,
+        test_isexecutable,
+        test_isreadable,
+        test_iswritable,
+        # test_chown,   # chmod & chown don't make sense for S3Paths
+        # test_chmod,
+    ]
+
+    # Run all of the automated tests
+    test(ps, testsets)
```
oxinabox: Oh awesome.

test/s3path.jl
```diff
@@ -0,0 +1,115 @@
...
+    root / "bar" / "qux/",
+    root / "bar" / "qux" / "quux.tar.gz",
+    root / "fred/",
+    root / "fred" / "plugh",
+    false
+)
+
+function test_s3_mkdir(p::PathSet)
+    @testset "mkdir" begin
+        garply = p.root / "corge" / "grault" / "garply/"
```
nickrobinson251: some terrific naming here 😆

test/s3path.jl
```diff
@@ -0,0 +1,115 @@
...
+        test_isreadable,
+        test_iswritable,
+        # test_chown,   # chmod & chown don't make sense for S3Paths
+        # test_chmod,
+    ]
+
+    # Run all of the automated tests
+    test(ps, testsets)
+
+    # TODO: Copy over specific tests that can't be tested reliably from the general case.
```
nickrobinson251: 🆗

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+    @assert startswith(str, "s3://")
+    tokenized = split(str, "/")
+    bucket = strip(tokenized[3], '/')
+    drive = "s3://$bucket"
+
+    if length(tokenized) > 3
+        root = "/"
+        # If the last tokenized element is an empty string then we've parsed a directory
+        dir = isempty(last(tokenized))
+        path = tuple(filter!(!isempty, tokenized[4:end])...)
```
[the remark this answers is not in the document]
oxinabox: It depends on Julia version.
Annoyingly. In 1.0 `tuple(x...)`, in 1.1 `Tuple(x)`

If one really cares workout how to do it  uses `ntuple`.

Project.toml
rofinn: That'll come when I tag a new release

src/s3path.jl
rofinn: 👌

src/s3path.jl
rofinn: Hmmm, true. Ideally, since we control the type we shouldn't be too worried about overriding the field.

Project.toml
oxinabox: I don't understand

src/s3path.jl
rofinn: Maybe? We were initially doing this a bunch before the `dir`/`isdirectory` field was introduced. I'm fine to leave it with the 2 cases we're doing it, but we might want to update that in the future.

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+
+FilePathsBase.ispathtype(::Type{S3Path}, str::AbstractString) = startswith(str, "s3://")
+FilePathsBase.isabs(fp::S3Path) = !isempty(fp.drive) && !isempty(fp.root)
+FilePathsBase.expanduser(fp::S3Path) = fp
+FilePathsBase.exists(fp::S3Path) = s3_exists(fp.config, fp.bucket, fp.key)
+Base.isfile(fp::S3Path) = !fp.dir && exists(fp)
+function FilePathsBase.isdir(fp::S3Path)
+    # Note: objects "s3://bucket/a" and "s3://bucket/a/b" can co-exist. If both of these
+    # objects exist listing the keys for "s3://bucket/a" returns ["s3://bucket/a"] while
+    # "s3://bucket/a/" returns ["s3://bucket/a/b"].
```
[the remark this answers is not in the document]
rofinn: Sure. I was going to leave a comment on the weirdness of directories on S3 in the constructor docstring.

src/s3path.jl
```diff
@@ -0,0 +1,235 @@
...
+Base.islink(fp::S3Path) = false
+Base.issocket(fp::S3Path) = false
+Base.isfifo(fp::S3Path) = false
+Base.ischardev(fp::S3Path) = false
+Base.isblockdev(fp::S3Path) = false
+Base.ismount(fp::S3Path) = false
+# Need API for accessing object ACL permissions for this to work
+FilePathsBase.isexecutable(fp::S3Path) = false
+Base.isreadable(fp::S3Path) = true
+Base.iswritable(fp::S3Path) = true
```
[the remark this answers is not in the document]
rofinn: Yes, and query the object ACL rules which isn't currently supported by AWSS3. I figured it would be better to work on that in a separate PR.

src/s3path.jl
rofinn: I didn't realize `read(path, String)` was being deprecated.

test/runtests.jl
```diff
@@ -258,31 +260,32 @@ versions = s3_list_versions(aws, bucket_name, "key3")
...
-end,
-enumerate(objs);
-ntasks=30)
+#     @test o == x[2]
+# end,
+# enumerate(objs);
+# ntasks=30)
 
-HTTP.ConnectionPool.showpool(stdout)
+# HTTP.ConnectionPool.showpool(stdout)
```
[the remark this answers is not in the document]
rofinn: They seemed to cause tests to stall? I'll re-enable when I'm ready to remove the WIP, but for now it makes testing locally difficult.

Project.toml
rofinn: The compat will be the next version of FilePathsBase which I haven't released yet.

src/s3path.jl
rofinn: NOTE: This is also an issue with FilePathsBase. https://github.com/rofinn/FilePathsBase.jl/issues/19

src/s3path.jl
oxinabox: I think it is gone

src/s3path.jl
rofinn: Oops, we don't have pushed docs setup for FilePathsBase yet.

src/s3path.jl
rofinn: It's the same as `Base.isdir`. If `FilePathsBase` is overloading `Base.isdir` does it matter?

src/s3path.jl
rofinn: 👌

Project.toml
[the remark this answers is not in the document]
rofinn: Compat for FilePathsBase has been added. I'm reluctant to add a new restriction for HTTP in this PR.

src/s3path.jl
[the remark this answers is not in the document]
rofinn: It is not listed in the notes for the `S3Path` docstring.

test/runtests.jl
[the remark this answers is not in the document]
rofinn: Hmmm, I was unable to figure out why these tests were stalling, so I've left a comment explaining that they seem to stall. I can try figuring out what's happening, but that's unrelated to this PR.

rofinn:
Alright, I think I've addressed most of the comments. I still don't understand why the unrelated asyncmap tests stall for me locally, but I let them run for 30 min and they never finished. I'm also unclear what they're supposed to be testing as object creation is handled elsewhere, so this is basically testing HTTP.jl?

src/AWSS3.jl
```diff
@@ -471,52 +474,63 @@ end
...
 
 [List Objects](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html)
 in `bucket` with optional `path_prefix`.
 
-Returns `Vector{Dict}` with keys `Key`, `LastModified`, `ETag`, `Size`,
+Returns an iterator of `Dict`s with keys `Key`, `LastModified`, `ETag`, `Size`,
 `Owner`, `StorageClass`.
 """
-function s3_list_objects(aws::AWSConfig, bucket, path_prefix="")
+function s3_list_objects(aws::AWSConfig, bucket, path_prefix=""; delimiter="/", max_items=nothing)
```
omus: Should the `delimiter` be a `Char`?

src/AWSS3.jl
```diff
@@ -471,52 +474,63 @@ end
...
+            q = Dict{String, String}()
+            if path_prefix != ""
+                q["prefix"] = path_prefix
+            end
+            if delimiter != ""
+                q["delimiter"] = delimiter
+            end
+            if marker != ""
+                q["marker"] = marker
+            end
```
omus: Personal preference but this may be easier to read:
```julia
!isempty(path_prefix) && (q["prefix"] = path_prefix)
!isempty(delimiter) && (q["delimiter"] = delimiter)
!isempty(marker) && (q["marker"] = marker)
```

src/s3path.jl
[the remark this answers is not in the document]
omus: @oxinabox what problem are you referring to? `Tuple` has always taken an iterator.

src/s3path.jl
```diff
@@ -0,0 +1,278 @@
...
+
+function Base.getproperty(fp::S3Path, attr::Symbol)
+    if attr === :anchor
+        return fp.drive * fp.root
+    elseif attr === :separator
+        return "/"
+    elseif attr === :bucket
+        return split(fp.drive, "//")[2]
+    elseif attr === :key
+        return fp.isdirectory ? join(fp.segments, '/') * "/" : join(fp.segments, '/')
```
omus:
```suggestion
        return join(fp.segments, '/') * (fp.isdirectory ? "/" : "")
```

src/s3path.jl
```diff
@@ -0,0 +1,278 @@
...
+# We need to special case join and parents so that we propagate
+# directories correctly (see type docstring for details)
+function Base.join(prefix::S3Path, pieces::AbstractString...)
+    isempty(pieces) && return prefix
+
+    segments = String[prefix.segments...]
+    isdirectory = endswith(last(pieces), "/")
+
+    for p in pieces
+        push!(segments, filter!(!isempty, split(p, "/"))...)
```
omus:
```suggestion
        append!(segments, filter!(!isempty, split(p, "/")))
```

src/s3path.jl
```diff
@@ -0,0 +1,278 @@
...
+        return getfield(fp, attr)
+    end
+end
+
+# We need to special case join and parents so that we propagate
+# directories correctly (see type docstring for details)
+function Base.join(prefix::S3Path, pieces::AbstractString...)
+    isempty(pieces) && return prefix
+
+    segments = String[prefix.segments...]
```
omus:
```suggestion
    segments = Vector{String}(prefix.segments)
```

src/s3path.jl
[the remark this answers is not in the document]
omus: Can we avoid defining these methods until then?

src/s3path.jl
```diff
@@ -0,0 +1,278 @@
...
+
+    @debug "delete: $fp"
+    s3_delete(fp.config, fp.bucket, fp.key)
+end
+
+# We need to special case sync with S3Paths because of how directories
+# are handled again.
+function FilePathsBase.sync(src::AbstractPath, dst::S3Path; delete=false)
+    # Create an index of all of the source files
+    index = Dict(Tuple(setdiff(p.segments, src.segments)) => p for p in walkpath(src))
```
omus: Does this always work?
```julia
julia> setdiff(["a", "b", "c"], ["b", "a", "d"])
1-element Array{String,1}:
 "c"
```

src/s3path.jl
```diff
@@ -0,0 +1,278 @@
...
+end
+
+function Base.readdir(fp::S3Path)
+    if isdir(fp)
+        k = fp.key
+        # Only list the files and "dirs" within this S3 "dir"
+        objects = s3_list_objects(fp.config, fp.bucket, k; delimiter="")
+
+        # Only list the basename and not the full key
+        basenames = unique(s3_get_name(k, string(o["Key"])) for o in objects)
```
omus:
```suggestion
        basenames = unique!(s3_get_name(k, string(o["Key"])) for o in objects)
```

test/runtests.jl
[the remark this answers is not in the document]
omus: Would have been better to use a comment block `#= =#` in this case

test/s3path.jl
```diff
@@ -0,0 +1,202 @@
+root = Path("s3://$bucket_name/pathset-root/")
+ps = PathSet(
+    root,
+    root / "foo/",
```
omus: Just curious if `root / "foo" /` works?

omus: Thinking about this more I realized we `/` to be supported as a post-fix operator.

src/AWSS3.jl
rofinn: I'd tried to stick with the original convention in this function. I tend to agree with your preferred syntax, but I think we should change the convention everywhere in 1 PR.

src/s3path.jl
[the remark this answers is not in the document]
oxinabox: @omus What is more performant between splatting + `tuple` and not splatting and `Tuple` varies between julia versions.
It is really weird but both Rory and I have done benchmarking on this.

src/s3path.jl
oxinabox: Can't call `unique!` on a generator.

src/AWSS3.jl
rofinn: I don't think so because we're passing it as a string to S3 anyways.

src/s3path.jl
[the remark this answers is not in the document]
rofinn: If folks find the `Tuple` syntax cleaner I'm fine to switch it. I don't think we care too much about performance here.

src/s3path.jl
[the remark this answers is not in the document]
rofinn: I'd prefer to keep it under this comment as a reminder of which methods to update. I suppose we could comment it out, but I don't think it makes a big difference. It'll just error anyways.

src/s3path.jl
rofinn: Yes? We can trust that `src.segments` is always a direct subset of `p.segments` based on how `walkpath` works. Is there something else in your example that I'm supposed to be noticing?

src/s3path.jl
rofinn: ☝️

src/s3path.jl
omus: Hmm, I see. Seems to be a slight abuse of `setdiff` but this should be ok

src/s3path.jl
omus: Don't use a generator?
```julia
julia> @btime unique(x for x in 1:100);
  2.394 μs (17 allocations: 6.17 KiB)

julia> @btime unique!([x for x in 1:100]);
  412.206 ns (1 allocation: 896 bytes)
```

oxinabox: feels like worth openning an upstream issue on.

Issue #51, opened 2019-10-15
Should `read(::S3Path)` return a `Vector{UInt8}`?


morris25:
`read` for  SystemPath, IO and String all return a `Vector{UInt8}`, while `read(::S3Path)` returns a `String`. Should the S3Path behavior be changed? Strings would still be accessible through `read(::S3Path, String)`.

morris25:
Added a fix, feel free to close both if you want to keep the current behavior.

PR #52, opened 2019-10-15
Unifies read output type with other Paths

morris25:
Makes `read` return a `Vector{UInt8}` to match SystemPaths.

Fixes https://github.com/JuliaCloud/AWSS3.jl/issues/51


mattBrzezinski:
This change makes sense to me, I don't see why anyone would object.

Issue #62, opened 2019-11-15
Cannot sync() files which are in directories that have the same name subsequently


mattBrzezinski:
Julia: `v1.2.0`
AWSS3.jl: `v0.6.4`

When you have a file on S3 in a directory with the same subsequent name doing a `readdir()` is fine, however a `sync()` call will fail with the following error:

> ERROR: LoadError: LoadError: ArgumentError: Source path does not exist: s3://mattbr-sample-s3-bucket/subdir1/

Below is a sample script to replicate this issue:

```
using Test

bucket_name = "mattbr-sample-s3-bucket"
file_name = "subdir1/subdir1/test.txt"
s3_create_bucket(bucket_name)
s3_put(bucket_name, file_name, "test")

@test readdir(S3Path("s3://$bucket_name/subdir1/subdir1/")) == ["test.txt"]

temp_path = p"temp-directory"
mkdir(temp_path)
#@test_throws ArgumentError sync(S3Path("s3://$bucket_name/"), temp_path)
sync(S3Path("s3://$bucket_name/"), temp_path)

#rm(temp_path, force=true, recursive=true)
```

Issue #61, opened 2019-11-15
Refactor testing


mattBrzezinski:
Current state of this package has tests all in one file, not real structure to them and dependent on previous tests.

The testing here needs to be refactored and re-structured.

PR #63, opened 2019-11-15
Fixed issue getting files from root S3 directory

### LINKED ISSUE #60: Most directory operatations do not work on the bucket itself, only on directory with in it
E.g. my bucket has a directory called `images` in it and a file also.
But `readdir` shows nothing
```
julia> using AWSS3, FilePathsBase

julia> readdir(S3Path("s3://oxinaboxpublic/"))
0-element Array{SubString{String},1}

julia> readdir(S3Path("s3://oxinaboxpublic/images/"))
4-element Array{SubString{String},1}:
 "chainloop.svg"
 "chainwithborder.png"
 "chainwithborder.svg"
 "chainwithborder_grey.png"
```

And `sync` fails:
```
julia> sync(S3Path("s3://oxinaboxpublic/images/"), p"out")

julia> readdir(p"out")
4-element Array{String,1}:
 "chainloop.svg"
 "chainwithborder.png"
 "chainwithborder.svg"
 "chainwithborder_grey.png"

julia> mkdir("outall")
"outall"

julia> sync(S3Path("s3://oxinaboxpublic/"), p"outall")
ERROR: ArgumentError: s3://oxinaboxpublic// does not exist
Stacktrace:
 [1] sync(::typeof(FilePathsBase.should_sync), ::S3Path, ::PosixPath; delete::Bool, overwrite::Bool) at /Users/
oxinabox/.julia/packages/FilePathsBase/Oyg1p/src/path.jl:537
 [2] sync at /Users/oxinabox/.julia/packages/FilePathsBase/Oyg1p/src/path.jl:537 [inlined]
 [3] #sync#16 at /Users/oxinabox/.julia/packages/FilePathsBase/Oyg1p/src/path.jl:532 [inlined]
 [4] sync(::S3Path, ::PosixPath) at /Users/oxinabox/.julia/packages/FilePathsBase/Oyg1p/src/path.jl:532
 [5] top-level scope at REPL[16]:1
```

I think something is wrong I guess with how we are interpretting prefixes as directories
Related to #34 maybe

mattBrzezinski:
Resolves #60 

There is a bug with reading files from the root "directory" of an S3 bucket, where when we access the `:key` attribute we would always return back `/` as a prefix. Files cannot be prefixed with this character as it's invalid, and thus we would always return back nothing.

This CR resolves the issue by check if the `fp.segments` are empty, if so return an empty string. `fp.segments` are used for "subdirectories" on an S3 bucket, for instance in the hierarchy below the `segments` for `file.txt` would be `[subdir1, subdir2]`. 

```
bucket-name
|-- subdir1
|     |-- subdir2
|          |-- file.txt
```

I have included testing for both `readdir()` for both S3 and local storage testing. One thing to note with local testing is the difference between an `S3Path` and `AbstractPath`. An `S3Path` will suffix a directory with the `/` character, while an `AbstractPath` will not. This has led to a special case when testing, if anyone has any suggestions on how to improve this I'd love to hear them!

Another thing which I have noticed when using `sync()` if you have two directories consecutively named the same thing, the files in the second directory will not be transferred and you will encounter an error with the `sync()` call. I have logged an issue for this found here: #62 

I have also moved the empty testing buckets code to the end of testing, so that you do not always have 1 bucket of testing data living in S3. I also have logged an issue to refactor testing #61 and just kept the current format going here.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -14,6 +14,7 @@ using HTTP
 using FilePathsBase
 using FilePathsBase.TestPaths
 using LinearAlgebra  # for norm S3Path tests
+using UUIDs: uuid4
 
 AWSCore.set_debug_level(1)
 
@@ -30,7 +31,6 @@ function test_without_catch(f)
 end
 
 
-
 #-------------------------------------------------------------------------------
 # Load credentials...
 #-------------------------------------------------------------------------------
@@ -44,31 +44,8 @@ aws[:region] = "ap-southeast-2"
 # S3 tests
 #-------------------------------------------------------------------------------
 
-# Delete old test files...
-
-for b in s3_list_buckets()
-
-    if occursin(r"^ocaws.jl.test", b)
-        @protected try
-            println("Cleaning up old test bucket: " * b)
-            @sync for v in s3_list_versions(aws, b)
-                @async s3_delete(aws, b, v["Key"]; version = v["VersionId"])
-            end
-            s3_delete_bucket(aws, b)
-        catch e
-            @ignore if isa(e, AWSCore.AWSException) &&
-                       e.code == "NoSuchBucket" end
-        end
-    end
-end
-
-HTTP.ConnectionPool.showpool(stdout)
-
 # Temporary bucket name...
-
-bucket_name = "ocaws.jl.test." * lowercase(Dates.format(now(Dates.UTC),
-                                                        "yyyymmddTHHMMSSZ"))
-
+bucket_name = "ocaws.jl.test." * lowercase(Dates.format(now(Dates.UTC), "yyyymmddTHHMMSSZ"))
 
 # Test exception code for deleting non existent bucket...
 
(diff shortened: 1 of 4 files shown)


test/issue_60.jl
```diff
@@ -0,0 +1,60 @@
+@testset "Issue 60" begin
```
nicoleepp: Can this testset and file get a more descriptive name than "issue 60"

test/runtests.jl
```diff
@@ -14,6 +14,7 @@ using HTTP
 using FilePathsBase
 using FilePathsBase.TestPaths
 using LinearAlgebra  # for norm S3Path tests
+using UUIDs
```
nicoleepp:
```suggestion
using UUIDs: uuid4
```

test/issue_60.jl
```diff
@@ -0,0 +1,60 @@
...
+        @test readdir(path / "subdir1/subdir2/subdir3/") == []
+    end
+
+    initialize()
+
+    @testset "readdir() - S3" begin
+        verify_files(S3Path("s3://$bucket_name/"))
+    end
+
+     @testset "readdir() - Local" begin
```
nicoleepp:
```suggestion
    @testset "readdir() - Local" begin
```

test/runtests.jl
```diff
@@ -290,6 +267,26 @@ versions = s3_list_versions(aws, bucket_name, "key3")
 
 
 include("s3path.jl")
+include("issue_60.jl")
```
nicoleepp:
```suggestion
```

oxinabox:
do not forget to bump the version number and tag a release

mattBrzezinski:
Forgot this package doesn't have `bors` setup, I'll leave this until Monday.

PR #64, opened 2019-12-04
added ListObjectsV2 call

ExpandingMan:
This adds the [ListObjectsV2](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html) call.  As far as I can tell, this is the only way to actually list all objects in a bucket.  No matter how I try, the older version never lists more than 1000 objects for me.

Right now the way I have this written gives lots of code duplication.  I left it this way for now because merging `s3_list_objects` and `s3_list_objects_v2` was a little bit messy.  As I see it we have 3 options:
- Leave it as is (two completely separate functions).
- Merge the two functions and handle all possible keys in one function.
- Deprecate `s3_list_objects` in favor of the new version.  I can't see any reason not to do this as the older version seems like a subset of the new one.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -538,8 +538,70 @@ function s3_list_objects(aws::AWSConfig, bucket, path_prefix=""; delimiter="/", 
     end
 end
 
-s3_list_objects(a...) = s3_list_objects(default_aws_config(), a...)
+s3_list_objects(a...; kwargs...) = s3_list_objects(default_aws_config(), a...; kwargs...)
 
+"""
+    s3_list_objects_v2([::AWSConfig], bucket, [path_prefix]; delimiter="/", max_items=1000)
+
+[List Objects](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html)
+in `bucket` with optional `path_prefix`.
+
+Returns an iterator of `Dict`s with keys `Key`, `LastModified`, `ETag`, `Size`,
+`Owner`, `StorageClass`.
+
+This uses the `ListObjectV2` function call.
+"""
+function s3_list_objects_v2(aws::AWSS3.AWSConfig, bucket, path_prefix=""; delimiter="/", max_items=nothing)
+    return Channel() do chnl
+        more = true
+        num_objects = 0
+        contoken = ""
+
+        while more
+            q = Dict{String, String}("list-type"=>"2")
+            if path_prefix != ""
+                q["prefix"] = path_prefix
+            end
+            if delimiter != ""
+                q["delimiter"] = delimiter
+            end
+            if contoken ≠ ""
+                q["continuation-token"] = contoken
+            end
+            if max_items !== nothing
+                q["max-keys"] = string(max_items - num_objects)
+            end
+
+            @repeat 4 try
+                # Request objects
+                r = s3(aws, "GET", bucket; query = q)
+
+                # Add each object from the response and update our object count / marker
+                if haskey(r, "Contents")
+                    l = isa(r["Contents"], Vector) ? r["Contents"] : [r["Contents"]]
+                    for object in l
+                        put!(chnl, xml_dict(object))
+                        num_objects += 1
+                    end
+                # It's possible that the response doesn't have "Contents" and just has a prefix,
+                # in which case we should just save the next marker and iterate.
+                elseif haskey(r, "Prefix")
+                    put!(chnl, Dict("Key" => r["Prefix"]))
+                    num_objects += 1
+                end
+
+                contoken = get(r, "NextContinuationToken", "")
+
+                # Continue looping if the results were truncated and we haven't exceeded out max_items (if specified)
+                more = r["IsTruncated"] == "true" && (max_items === nothing || num_objects < max_items)
+            catch e
+                @delay_retry if ecode(e) in ["NoSuchBucket"] end
+            end
+        end
+    end
+end
+
+s3_list_objects_v2(a...; kwargs...) = s3_list_objects_v2(default_aws_config(), a...; kwargs...)
 
 """
     s3_list_keys([::AWSConfig], bucket, [path_prefix])


bkamins:
CC @pszufe

mattBrzezinski:
> No matter how I try, the older version never lists more than 1000 objects for me.

[ListObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html) is limited to 1,000 objects.

> Deprecate s3_list_objects in favor of the new version. I can't see any reason not to do this as the older version seems like a subset of the new one.

I think this makes the most sense.

src/AWSS3.jl
```diff
@@ -538,8 +538,70 @@ function s3_list_objects(aws::AWSConfig, bucket, path_prefix=""; delimiter="/",
...
+
+        while more
+            q = Dict{String, String}("list-type"=>"2")
+            if path_prefix != ""
+                q["prefix"] = path_prefix
+            end
+            if delimiter != ""
+                q["delimiter"] = delimiter
+            end
+            if contoken ≠ ""
```
mattBrzezinski: For consistency we should probably keep using `!=` instead of `≠`

src/AWSS3.jl
```diff
@@ -538,8 +538,70 @@ function s3_list_objects(aws::AWSConfig, bucket, path_prefix=""; delimiter="/",
...
+            end
+
+            @repeat 4 try
+                # Request objects
+                r = s3(aws, "GET", bucket; query = q)
+
+                # Add each object from the response and update our object count / marker
+                if haskey(r, "Contents")
+                    l = isa(r["Contents"], Vector) ? r["Contents"] : [r["Contents"]]
+                    for object in l
```
mattBrzezinski: Instead of iterating over these objects, could we not just `put!` the vector into the channel? (Not sure if this is possible).

And then just do `num_objects += size(l)[1]`

pszufe:
In production systems an S3 bucket can contain millions of items and S3 has no directory structure (this what you see as directories are in fact name prefixes so there is not such thing as directory index). 
Maybe since you have a chance yo use a new AWS API it would be also a good idea to update the AWSS3 API? 
Perhaps, `s3_list_objects_v2`  could return some `ResultSet`  obejct and then one can call `getdata(resultset)` and `next!(resultset)`. Otherwise for large scale system this API will be not usable anyway (listing 1million objects can take very considerable amount of time while a typical large scale scenario is to list objects and asynchronously launch jobs for processing them).

ExpandingMan:
Note that there is (and has been) an option for limiting the number of items returned, so I'd assume it would work fine as is.

Is there a consensus on deprecating the older `ListObjects` in favor of this new `ListObjectsV2`?

If so I will eliminate the old function and try to make sure we don't lose any functionality in the new one (might have to add a couple of keys, I will check).

pszufe:
The AWS S# REST API docs reads (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html):
> We recommend that you use this revised API for application development. For backward compatibility, Amazon S3 continues to support the prior version of this API, ListObjects. 

So it seems to be reasonable to depreciate the old version.

mattBrzezinski:
I am in favour of adding the `@deprecated` tag to the `ListObjects` function, introducing the `ListObjectsV2` call and releasing a patch rather than removing the `ListObjects` function and doing a major release.

ExpandingMan:
Any concensus about what should be done here so I can update the PR?  Are we ok with deprecating the old version?

s2maki:
The only reason I can think of for keeping both is that the pagination arguments are different between calls.  ListObjects uses `Marker`, while V2 uses `ContinuationToken`.  I use ListObjects, but I expect my object count to be small (a couple dozen).  Anything more than that, and I find I have to maintain my index another way anyhow.

mattBrzezinski:
X-posting a similar merge request from @pschmied back in here:

> https://github.com/JuliaCloud/AWSS3.jl/pull/94

pschmied:
> The only reason I can think of for keeping both is that the pagination arguments are different between calls. ListObjects uses `Marker`, while V2 uses `ContinuationToken`. I use ListObjects, but I expect my object count to be small (a couple dozen). Anything more than that, and I find I have to maintain my index another way anyhow.

In the V1 ListObjects API, the "marker" parameter causes the API to return objects that follow the marker lexicographically:

https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html#API_ListObjects_RequestSyntax

In the ListObjectsV2 API, the "start-after" parameter causes the API to return objects that follow the start-after parameter lexicographically:

https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_RequestSyntax


My feeling is that exposing the "start-after" parameter in the s3_list_objects_v2 function would allow anyone relying on the old "marker" parameter to do what they need to do. Thus you could deprecate the old ListObjects / s3_list_objects. Do others agree?

src/AWSS3.jl
pschmied: Wouldn't the xml_dict() function still need to be applied to each element in that vector before `put!`-ing the vector to the channel?

src/AWSS3.jl
```diff
@@ -538,8 +538,70 @@ function s3_list_objects(aws::AWSConfig, bucket, path_prefix=""; delimiter="/",
...
+
+[List Objects](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html)
+in `bucket` with optional `path_prefix`.
+
+Returns an iterator of `Dict`s with keys `Key`, `LastModified`, `ETag`, `Size`,
+`Owner`, `StorageClass`.
+
+This uses the `ListObjectV2` function call.
+"""
+function s3_list_objects_v2(aws::AWSS3.AWSConfig, bucket, path_prefix=""; delimiter="/", max_items=nothing)
```
pschmied, line 554:
```suggestion
function s3_list_objects_v2(aws::AWSS3.AWSConfig, bucket, path_prefix=""; delimiter="/", start_after="", max_items=nothing)
```

src/AWSS3.jl
```diff
@@ -538,8 +538,70 @@ function s3_list_objects(aws::AWSConfig, bucket, path_prefix=""; delimiter="/",
...
+            q = Dict{String, String}("list-type"=>"2")
+            if path_prefix != ""
+                q["prefix"] = path_prefix
+            end
+            if delimiter != ""
+                q["delimiter"] = delimiter
+            end
+            if contoken ≠ ""
+                q["continuation-token"] = contoken
+            end
```
pschmied, line 570:
```suggestion
            end
            if start_after != ""
                q["start-after"] = start_after
            end
```

src/AWSS3.jl
pschmied, line 568:
```suggestion
            if contoken != ""
```

pschmied:
Just bumping here to see if there's anything more I can do to help move this along.

pschmied:
If it is useful, I have collected I think all of the feedback, plus tests into the master branch of my fork: https://github.com/pschmied/AWSS3.jl

mattBrzezinski:
> Just bumping here to see if there's anything more I can do to help move this along.

I'm currently quite busy with other work, but at some point I will have time to dedicate to this. If someone else could take the lead on reviewing / giving feedback that'd be greatly appreciated!

PR #70, opened 2020-02-14
Fix typo in error message 's'->'s3'

nickrobinson251:


test/s3path.jl
```diff
@@ -252,6 +252,7 @@ end
 
     @testset "S3" begin
         verify_files(S3Path("s3://$bucket_name/"))
+        @test_throws ArgumentError("$bucket_name doesn't start with s3://") S3Path(bucket_name)
```
mattBrzezinski: Not sure how I feel about including the text that's thrown. I think just asserting that an `ArgumentError` is thrown is good.

The issue comes further down the line if we change it, we're essentially duplicating these strings. It would be nice if we wanted to be explicit to have an `errors.jl` file, just containing all the strings used throughout the project and just pull them from there. But that's definitely over kill for now and for this CR.

nickrobinson251: 🤷‍♂ up to you

ideally we'd test that the error message contains the string "s3://" which is the crucial bit of info -- tbh an ArgumentError with including info about needing "s3://" is pretty useless here

but I don't think Test.@test_throws supports that?

mattBrzezinski: Oh sorry, I meant for just testing purposes. Unless `verify_files()` can throw a different `ArgumentError` message, it just seems like we're duplicating a bit of code here. But thinking about it some more there's no real point in refactoring it now since this is a one off thing. But if this starts becoming more common then we can consider it!

PR #74, opened 2020-03-24
Fix S3Path `join`

### LINKED ISSUE #71: Joining `S3Path` with `/` fails to add first directory split
```julia
julia> Path("foo") / "bar" / "baz"
p"foo/bar/baz"

# expected p"s3://foo/bar/baz"
julia> S3Path("s3://foo") / "bar" / "baz"
p"s3://foobar/baz"  
```

I think this is happening because it's just dispatching to `join`
```julia
julia> @which S3Path("s3://foo") / "bar" / "baz"
/(root::AbstractPath, pieces::Union{AbstractString, AbstractPath}...) in FilePathsBase at /Users/nick/.julia/packages/FilePathsBase/oi7XZ/src/path.jl:215
```
https://github.com/rofinn/FilePathsBase.jl/blob/v0.7.0/src/path.jl#L214-L216

and for `join` we identify S3 "directories" by the trailing `/`
```julia
julia> @which join(S3Path("s3://foo"), "bar", "baz")
join(prefix::S3Path, pieces::AbstractString...) in AWSS3 at /Users/nick/.julia/packages/AWSS3/VMuUL/src/s3path.jl:94
```
https://github.com/JuliaCloud/AWSS3.jl/blob/v0.6.8/src/s3path.jl#L93-L97

As the `S3Path` docstring explains, this is to do with S3 "faking" directories
https://github.com/JuliaCloud/AWSS3.jl/blob/v0.6.8/src/s3path.jl#L18-L23

But while the `join` behaviour is arguably correct (although i think it's arguable)
```julia
julia> join(S3Path("s3://foo"), "bar", "baz")
p"s3://foobar/baz"
```
it leads to surprising behaviour when using the `/` syntax.

I wonder whether is is worth special-casing the `/` syntax for `S3Paths` to behave like
```julia
julia> S3Path("s3://foo") / "bar" / "baz"
p"s3://foo/bar/baz"  
```

nickrobinson251:
- closes #71

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -102,7 +102,7 @@ function Base.join(prefix::S3Path, pieces::AbstractString...)
 
     return S3Path(
         tuple(segments...),
-        prefix.root,
+        "/",
         prefix.drive,
         isdirectory,
         prefix.config,
--- a/test/s3path.jl
+++ b/test/s3path.jl
@@ -263,7 +263,7 @@ end
         verify_files(temp_path)
 
         rm(temp_path, force=true, recursive=true)
-     end
+    end
 
     @testset "join" begin
         @test (  # test trailing slash on prefix does not matter for join
--- a/test/s3path.jl
+++ b/test/s3path.jl
@@ -278,7 +278,7 @@ end
         )
         # test extra leading and trailing slashes do not matter
         @test p"s3://foo/" / "bar/" / "/baz" == p"s3://foo/bar/baz"
-        # test joining `/` and string concatentation `*` play nice as expect
+        # test joining `/` and string concatentation `*` play nice as expected
         @test p"s3://foo" * "/" / "bar" == p"s3://foo" / "/" * "bar" == p"s3://foo" / "bar"
         @test p"s3://foo" / "bar" * "baz" == p"s3://foo/bar" * "baz"  == p"s3://foo" / "barbaz"
         # test trailing slash on final piece is included
--- a/test/s3path.jl
+++ b/test/s3path.jl
@@ -271,7 +271,7 @@ end
             p"s3://foo/bar/" / "baz" ==
             p"s3://foo/bar/baz"
         )
-        @test_broken (  # test trailing slash on root-only prefix in particular does not matter
+        @test (  # test trailing slash on root-only prefix in particular does not matter
             p"s3://foo" / "bar" / "baz" ==
             p"s3://foo/" / "bar" / "baz" ==
             p"s3://foo/bar/baz"
--- a/test/s3path.jl
+++ b/test/s3path.jl
@@ -278,6 +278,9 @@ end
         )
         # test extra leading and trailing slashes do not matter
         @test p"s3://foo/" / "bar/" / "/baz" == p"s3://foo/bar/baz"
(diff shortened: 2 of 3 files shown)


nickrobinson251:
fyi @rofinn @mattBrzezinski

nickrobinson251:
this now feels better to me:
```julia
julia> foo = p"s3://foo"
p"s3://foo"

julia> foo * "bar" / "baz"
p"s3://foobar/baz"

julia> foo / "bar" / "baz"  # before this PR, this was the same as above
p"s3://foo/bar/baz"
```
Should i document this in a test too?

nickrobinson251:
More generally, I couldn't see any docs that should be updated, but let me know if and where I should add some docs

src/s3path.jl
```diff
@@ -102,7 +102,7 @@ function Base.join(prefix::S3Path, pieces::AbstractString...)
 
     return S3Path(
         tuple(segments...),
-        prefix.root,
+        "/",
```
oxinabox: explain this change so I understand?

nickrobinson251: I'm not 100% here. but the thought is this:

if we start with `foo`:
```julia
julia> foo = p"s3://foo"
p"s3://foo"

julia> foo.root
""
```
and we `new_foobar = join(foo, "bar")` correctly, so as to get it `== foobar`:
```julia
julia> foobar = p"s3://foo/bar"
p"s3://foo/bar"

julia> foobar.root
"/"
```
then we have to set the root of `new_foobar` to `"/"`

nickrobinson251: i.e. without this change (the thing this PR is fixing):
```julia
julia> new_foobar = join(foo, "bar")
p"s3://foobar"

julia> new_foobar.root
""

julia> awkward_foobar = join(foo * "/", "bar")
p"s3://foo/bar"

julia> awkward_foobar.root
"/"
```

oxinabox:
I think this needs @rofinn  to review.

mjram0s:
Looks like this solves the problem mentioned in the issue and tests look sufficient! Approved.

test/s3path.jl
```diff
@@ -263,5 +263,22 @@ end
         verify_files(temp_path)
 
         rm(temp_path, force=true, recursive=true)
-     end
+    end
+
```
mattBrzezinski: Could you just add in a couple tests w/ `*`? And then I think it's good to go, we don't seem to have any for that functionality in here.

nickrobinson251: added

test/s3path.jl
```diff
@@ -263,5 +263,25 @@ end
...
+            p"s3://foo/bar/baz"
+        )
+        @test (  # test trailing slash on root-only prefix in particular does not matter
+            p"s3://foo" / "bar" / "baz" ==
+            p"s3://foo/" / "bar" / "baz" ==
+            p"s3://foo/bar/baz"
+        )
+        # test extra leading and trailing slashes do not matter
+        @test p"s3://foo/" / "bar/" / "/baz" == p"s3://foo/bar/baz"
+        # test joining `/` and string concatentation `*` play nice as expect
```
nickrobinson251:
```suggestion
        # test joining `/` and string concatentation `*` play nice as expected
```

PR #78, opened 2020-04-03
Extra S3Path Constructors

rofinn:
In particular, if you explicitly want to construct an `S3Path` with a bucket, key and specify whether it's a directory. This should be more convenient than calling the rather low level default constructor or requiring some manual string manipulations.

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -64,7 +64,7 @@ function S3Path(
     return S3Path(
         key.segments,
         "/",
-        strip(startswith(bucket, "s3://") ? bucket : "s3://$bucket", '/'),
+        normalize_bucket_name(bucket),
         isdirectory,
         config,
     )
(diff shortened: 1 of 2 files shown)


src/s3path.jl
```diff
@@ -40,6 +40,36 @@ function S3Path()
...
+function S3Path(
+    bucket::AbstractString,
+    key::AbstractPath;
+    isdirectory::Bool=false,
+    config::AWSConfig=aws_config(),
+)
+    return S3Path(
+        key.segments,
+        "/",
+        strip(startswith(bucket, "s3://") ? bucket : "s3://$bucket", '/'),
```
oxinabox, line 67: maybe extract this into a function, as a way to give it a clear name.
`normalize_bucket_name(bucket)`

rofinn: Sure

PR #82, opened 2020-04-07
Test CI on LTS and latest release

omus:
We should be testing on more modern versions of Julia


omus:
I'll rebase this once the Compat PRs are merged

omus:
Newer versions of packages are being used on Julia 1.3. :)

PR #83, opened 2020-04-07
Remove unused DataStructures dependency

omus:
I was going to switch DataStructures to OrderedCollections when I realized that `OrderedDict` isn't even used in this package.


omus:
Note that DataStructures is still used by dependents so the only thing that has changed is that DataStructures cannot be used by this package.

Issue #84, opened 2020-04-11
s3_upload_path errors


Keno:
```
type Array has no field headers
Stacktrace:
 [1] getproperty(::Array{UInt8,1}, ::Symbol) at ./Base.jl:33
 [2] s3_upload_part(::Dict{Symbol,Any}, ::XMLDict.XMLDictElement, ::Int64, ::Array{UInt8,1}) at /home/keno/.julia/dev/AWSS3/src/AWSS3.jl:697
```

Looks like this function expects to get a raw response object back from the API, but instead only gets the body.

mattBrzezinski:
Looking into this I can see where this issue is stemming from.

```julia
function AWSS3.s3_upload_part(aws::AWSConfig, upload, part_number, part_data)
    response = s3(aws, "PUT", upload["Bucket"];
                  path = upload["Key"],
                  query = Dict("partNumber" => part_number,
                               "uploadId" => upload["UploadId"]),
                  content = part_data)

    response.headers["ETag"]  # Code breaks here
end
```

```julia
function AWSS3.s3()
    ....
    response = AWSCore.do_request(request)
    ....
```

```julia
function AWSCore.do_request(r::AWSRequest)
    ....
    return response.body
    ....
```

The only time headers get returned from `AWSCore` is if a `HEAD` request is performed, [code](https://github.com/JuliaCloud/AWSCore.jl/blob/master/src/AWSCore.jl#L498).

mattBrzezinski:
@Keno do you require the `ETag`, or any of the other [common response headers](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html)? If not I propose that we modify `s3_upload_part()` just return the `response` for now.

Alternatively, we can release a new version of `AWSCore` which `do_request()` returns back the full `response` object by default, and version bump `AWSS3` to match.

mattBrzezinski:
Resolved in #86

PR #86, opened 2020-04-15
Fixed S3 Multi-Part Upload

mattBrzezinski:
**Depends On:** https://github.com/JuliaCloud/AWSCore.jl/pull/117
**Resolves:** #84 

## Overview
Previously `AWSCore.jl` was not returning headers for a request and the `s3_upload_part()` would break. Changes have been made in that package to return response headers when the kwarg `return_headers=true` is passed in.

This merge request is just updating the `s3_upload_part()` to properly pull the returned `ETag` from Amazon, as well as include a test for multi-part uploading.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -59,7 +59,8 @@ function s3(aws::AWSConfig,
             version="",
             content="",
             return_stream=false,
-            return_raw=false,)
+            return_raw=false,
+            return_headers=false)
 
     # Build query string...
     if version != ""
@@ -106,7 +107,7 @@ function s3(aws::AWSConfig,
         end
         request[:url] = url
 
-        response = AWSCore.do_request(request)
+        response = AWSCore.do_request(request; return_headers=return_headers)
 
         # Handle 301 Moved Permanently with missing Location header.
         # https://github.com/samoconnor/AWSS3.jl/issues/25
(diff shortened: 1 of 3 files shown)


Keno:
I've been playing with this and it works fine for simple uploads. However, larger uploads quickly run into connection errors from HTTP.jl. A bunch of them are similar to https://github.com/JuliaWeb/HTTP.jl/issues/517, so I've spent some time debugging that, but unfortunately, that doesn't appear sufficient. Shouldn't block this merge of course.

Keno:
Ok, the other errors I ended up with were just expirations of the access tokens, though HTTP.jl really didnt't like that.

src/AWSS3.jl
```diff
@@ -106,7 +107,7 @@ function s3(aws::AWSConfig,
         end
         request[:url] = url
 
-        response = AWSCore.do_request(request)
+        response = AWSCore.do_request(request; return_headers=return_headers)
```
iamed2, line 110: This could return a tuple, which means you need to change the `isa` check below.

Maybe `(response isa XMLDict.XMLDictElement || response isa Tuple{XMLDict.XMLDictElement, Any})`?

mattBrzezinski: Could also do:

```julia
response = AWSCore.do_request(request; return_headers=return_headers)

if response isa Tuple{XMLDict.XMLDictElement, Any}
  response = response[1]
end

if response isa XMLDict.XMLDictElement && ...
```

or this way, which I like a lot less

```julia
if return_headers
  response, _ = AWSCore.do_request(request; return_headers=return_headers)
else
  response = AWSCore.do_request(request; return_headers=return_headers)
end

if response isa XMLDict.XMLDictElement && ...
```

iamed2: Well the first way discards the headers, but you need to return them later if `return_headers`

PR #88, opened 2020-05-01
Update to the latest version of FilePathsBase

rofinn:
- `tryparse` instead of `ispathtype/constructor`.
- Breaking improvements to `parent(s)` functions
- `/` and `join` need to be explicitly imported from FilePathsBase and don't extend Base
- Function renames that aren't explicitly implemented for S3Path, but required updating tests

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -70,9 +70,16 @@ function S3Path(
     )
 end
 
+# To avoid a breaking change.
 function S3Path(str::AbstractString; config::AWSConfig=aws_config())
+    result = tryparse(S3Path, str; config=config)
+    result !== nothing || throw(ArgumentError("Invalid s3 path string: $str"))
+    return result
+end
+
+function Base.tryparse(::Type{S3Path}, str::AbstractString; config::AWSConfig=aws_config())
     str = String(str)
-    startswith(str, "s3://") || throw(ArgumentError("$str doesn't start with s3://"))
+    startswith(str, "s3://") || return nothing
     root = ""
     path = ()
     isdirectory = true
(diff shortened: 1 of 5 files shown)


test/s3path.jl
```diff
@@ -159,27 +173,30 @@ end
...
         test_s3_join,
+        test_splitext,
         test_basename,
         test_filename,
         test_extensions,
         test_isempty,
-        test_s3_norm,
-        # test_real, # real doesn't make sense for S3Paths
+        test_s3_normalize,
+        # test_canonicalize, # real doesn't make sense for S3Paths
```
mattBrzezinski, line 193: If we're not testing this then we should probably just remove it.

src/s3path.jl
```diff
@@ -145,7 +152,7 @@ end
 
 function FilePathsBase.parents(fp::S3Path)
     if hasparent(fp)
-        return map(1:length(fp.segments)-1) do i
+        return map(0:length(fp.segments)-1) do i
```
mattBrzezinski, line 155: What's the reasoning for this change, just curious?

rofinn: Parents should include the root if we're dealing with an absolute path. In the case of S3Paths, they're always absolute paths, so this behaviour in FilePathsBase isn't necessary.

https://github.com/rofinn/FilePathsBase.jl/blob/master/src/path.jl#L224

test/s3path.jl
rofinn: I kept it there in case we want to check that `canonicalize(::S3Path)` is a no-op in the future.

PR #90, opened 2020-06-16
Add signature v4

kevin-yen:
*Changes*
* Added signature v4 pre-signed URL function
* Extracted out current v2 signature process to a separate function
* Added `signature_version` keyword parameter to `s3_sign_url` to select which signature process version to use
* Set default signature version to v4
* Updated documentation

PR #91, opened 2020-06-26
Version bump 0.7.1

mattBrzezinski:
Includes

- https://github.com/JuliaCloud/AWSS3.jl/pull/90


iamed2:
I noticed that this adds DataStructures when it only requires OrderedCollections. Up to you whether you want to adjust that.

PR #94, opened 2020-07-31
Update s3_list_objects() to use ListObjectsV2

pschmied:
Hello, I noticed that this library was still using the ListObjects API, which has been deprecated in favor of ListObjectsV2 (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html). I thought I would use this as an opportunity to dip my feet back into Julia after a few years. Style or other suggestions / corrections humbly welcomed.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -497,46 +497,58 @@ function s3_list_objects(aws::AWSConfig, bucket, path_prefix=""; delimiter="/", 
     return Channel() do chnl
         more = true
         num_objects = 0
-        marker = ""
+        next_token = ""
 
         while more
             q = Dict{String, String}()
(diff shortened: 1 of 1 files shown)


mattBrzezinski:
Hi Peter, welcome to back to Julia!

There's already a merge request opened up for this #64 and the discussion around it.

pschmied:
Oh, oops! Shame on me for failing to look at open pull requests! I'll take a look, but I suspect there is unlikely to be anything in my code that is better or even substantially different than what's there.

mattBrzezinski:
No worries! :)

I'm going to close this merge request, but I'm sure we can x-post this over and unify things if possible.

PR #96, opened 2020-08-10
Wait for entries to be available in `isdir` objects channel

rofinn:
The tests were failing locally on julia 1.5, but the CI hadn't run for a month. As a result, this is just a 1 line change with some comments explaining why we're using `iterate`.

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -174,7 +174,12 @@ function Base.isdir(fp::S3Path)
     end
 
     objects = s3_list_objects(fp.config, fp.bucket, key; max_items=1)
-    return !isempty(objects)
+
+    # `objects` is a `Channel`, so we call iterate to see if there are any objects that
+    # match our directory key.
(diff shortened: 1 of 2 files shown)


src/s3path.jl
```diff
@@ -174,7 +174,12 @@ function Base.isdir(fp::S3Path)
     end
 
     objects = s3_list_objects(fp.config, fp.bucket, key; max_items=1)
-    return !isempty(objects)
+
+    # `objects` is a `Channel`, so we call iterate to see if there are any objects that
```
mattBrzezinski, line 178: Would `iterate` not be checking the first item and returning a tuple of itself and the init state or `nothing`?

> https://docs.julialang.org/en/v1/manual/interfaces/

rofinn: That's correct. Since this function just wants to know if there are any items that match we're just checking that it doesn't return `nothing`.

PR #107, opened 2020-10-14
Use AWS.jl instead of AWSCore.jl

mattBrzezinski:
## Overview
The purpose of this merge request is to replace [AWSCore.jl](https://github.com/JuliaCloud/AWSCore.jl) with the new shiny [AWS.jl](https://github.com/JuliaCloud/AWS.jl).

**This is a breaking change, since end users must use AWS.AWSConfig not AWSCore.AWSConfig structs**

Those with a sharp eye will notice that almost every definition uses the high-level S3 wrapper, except for a few cases. This is due to https://github.com/JuliaCloud/AWS.jl/issues/219 which is preventing RESTXML (protocol which S3 uses) to make PUT requests with body data properly. This should not be a blocking issue, as we can convert to the high-level functions when the issue is resolved.


mattBrzezinski:
bors try

oxinabox:
Is this nonbreaking for people who never used the config struct and always leave it as the default that is read off the filesystem?

mattBrzezinski:
> Is this nonbreaking for people who never use the config struct and always leave it as the default that is read off the filesystem?

That's correct.

I am however debating modifying the function signatures (since we're already making a breaking change) from:

```julia
function s3_list_objects(aws::AWSConfig, bucket, path_prefix=""; delimiter="/", max_items=nothing, kwargs...)
```

```julia
function s3_list_objects(bucket, path_prefix=""; aws::AWSConfig=global_aws_config(), delimiter="/", max_items=nothing, kwargs...)
```

sean-bennett112:
Just tested this for our(Beacon's) use cases, and it works well as a drop-in replacement for what we would use it for. :+1:

mattBrzezinski:
bors try

mattBrzezinski:
bors try

sean-bennett112:
Just looked it over again after recent commits. Still looks good, and confirmed that it still works to replace our internal repo. :shipit:

mattBrzezinski:
> Just looked it over again after recent commits. Still looks good, and confirmed that it still works to replace our internal repo. :shipit:

Ty! Sending it now. #119 will need a re-work, it's currently broken but looking into it.

PR #108, opened 2020-10-20
Leave proxy management to HTTP.jl

jademackay:
Remove proxy functionality from `s3` function since HTTP.jl handles this nicely using ENV.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -62,8 +62,7 @@ function s3(aws::AWSConfig,
             content="",
             return_stream=false,
             return_raw=false,
-            return_headers=false,
-            proxy=nothing)
+            return_headers=false)
 
     # Build query string...
     if version != ""
@@ -72,7 +71,6 @@ function s3(aws::AWSConfig,
 
     query_str = HTTP.escapeuri(query)
     resource = string("/", HTTP.escapepath(path), query_str == "" ? "" : "?$query_str")
-    http_options = @SymDict(proxy)
 
     # Build Request...
     request = @SymDict(service = "s3",
@@ -82,7 +80,6 @@ function s3(aws::AWSConfig,
                        content,
                        return_stream,
                        return_raw,
-                       http_options,
                        aws...)
 
     @repeat 3 try
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -82,6 +82,7 @@ function s3(aws::AWSConfig,
                        content,
                        return_stream,
                        return_raw,
+                       http_options,
                        aws...)
 
     @repeat 3 try
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -62,7 +62,8 @@ function s3(aws::AWSConfig,
             content="",
             return_stream=false,
             return_raw=false,
-            return_headers=false)
+            return_headers=false,
+            proxy=nothing)
 
     # Build query string...
     if version != ""
@@ -71,6 +72,7 @@ function s3(aws::AWSConfig,
 
     query_str = HTTP.escapeuri(query)
     resource = string("/", HTTP.escapepath(path), query_str == "" ? "" : "?$query_str")
+    http_options = proxy === nothing ? Dict{Symbol,Any}() : @SymDict(proxy)
 
     # Build Request...
     request = @SymDict(service = "s3",


mattBrzezinski:
If a user is not using `ENV` this still allows them to pass in their own proxy. This package is using [AWSCore.jl](https://github.com/JuliaCloud/AWSCore.jl/blob/2084eb2608d02660662a0c47956c70b08bee22f5/src/http.jl#L29) to make the requests which allows for passing along `http_options` to the request itself, and not [HTTP.jl](https://github.com/JuliaWeb/HTTP.jl) directly.

I'm not convinced that this functionality should be removed here, do you have any reasons why it should?

jademackay:
Yes, sorry, I should have provided some reasoning.
The current implementation *requires* the user to specify the proxy. This  means that the user must apply different code depending on whether they are behind a proxy or not. The behavior I prefer is not to have to specify the proxy yet preserve the ability to specify a proxy if desired. I think the simplest way to do this is to fall back on HTTP which uses ENV. An alternative I considered was
```
function s3(,...;proxy=get(ENV,"https_proxy",nothing),...)
```
however, this obscures `HTTP.jl`'s  `getproxy` logic
```
function getproxy(scheme, host)                                                           
    isnoproxy(host) && return nothing                                                     
    if scheme == "http" && haskey(ENV, "http_proxy")                                      
        return ENV["http_proxy"]                                                          
    elseif scheme == "https" && haskey(ENV, "https_proxy")                                
        return ENV["https_proxy"]                                                         
    end                                                                                   
    return nothing                                                                        
    end
```
where `host` is the the url host (something like `amasonaws.$(region).s3.....com`).

mattBrzezinski:
> The current implementation _requires_ the user to specify the proxy. This means that the user must apply different code depending on whether they are behind a proxy or not. The behavior I prefer is not to have to specify the proxy yet preserve the ability to specify a proxy if desired. I think the simplest way to do this is to fall back on HTTP which uses ENV. An alternative I considered was
> 
> ```
> function s3(,...;proxy=get(ENV,"https_proxy",nothing),...)
> ```

The `proxy` kwarg is available in all functions, which defaults to `nothing`. If you do not want to specify a `proxy` you do not need to pass anything in here just use the functionality as intended.

If you're always behind a proxy and need to pass in the kwarg you can just do:

```julia
using AWSS3
conf = default_aws_config()

bucket_name = "example-bucket"
obj_name = "example-obj"
kwargs = Dict(:proxy=>true)

s3_create_bucket(conf, bucket_name; kwargs...)
s3_put(conf, bucket_name, obj_name; kwargs...)
s3_get(conf, bucket_name, obj_name; kwargs...)
```

I have not looked at `HTTP.jl` in a long time, but this shouldn't interfere with the `getproxy()` functionality, just set the ENV VAR, and don't pass in a `proxy` and it should work itself out?

Every request this package makes gets passed through here, off to HTTP.jl,
https://github.com/JuliaCloud/AWSCore.jl/blob/2084eb2608d02660662a0c47956c70b08bee22f5/src/http.jl#L42-L54

jademackay:
I understand that `proxy=nothing` (the default value) currently behaves more like what you would expect from`proxy=false`.

My issue is that I am not always behind a proxy. For code that depends on `HTTP.jl` this is not a problem, however for `AWSS3.jl` it is. I suppose I  could create `kwargs = Dict(:proxy=>get(ENV,"https_proxy",nothing)` and add `;kwargs...) to every AWS function call but I ideally prefer not to do this kind of wrapping. 

Perhaps I should look again again `AWSCore.jl`'s `http.jl` file. My feeling is that `proxy=nothing` should fallback to `HTTP.jl`'s behavior and `proxy=myproxy` should be honored.

jademackay:
Oops, the `proxy=myproxy` is not honored but eliminated in the above

mattBrzezinski:
> I understand that `proxy=nothing` (the default value) currently behaves more like what you would expect from`proxy=false`.
> 
> My issue is that I am not always behind a proxy. For code that depends on `HTTP.jl` this is not a problem, however for `AWSS3.jl` it is. I suppose I could create `kwargs = Dict(:proxy=>get(ENV,"https_proxy",nothing)` and add `;kwargs...) to every AWS function call but I ideally prefer not to do this kind of wrapping.
> 
> Perhaps I should look again again `AWSCore.jl`'s `http.jl` file. My feeling is that `proxy=nothing` should fallback to `HTTP.jl`'s behavior and `proxy=myproxy` should be honored. I think this could be achieved by adding
> 
> ```
> ENV["https_proxy"] = myproxy
> ```
> 
> with the proxy kwargs removed as in the PR, instead of
> 
> ```
> kwargs = Dict(:proxy=>myproxy)
> 
> s3_create_bucket(conf, bucket_name; kwargs...)
> ...
> ```

I think we can work around this by doing a check if `proxy !=== nothing` and then creating the `http_options` and passing that along to actually making a request.

One reason I don't want to remove this now is that it's a breaking change and requires yet another version bump. If you're using `v0.7.3` or earlier with a `proxy` does the underlying request go through and `HTTP.jl` handles the proxy?

jademackay:
Yes, that would be the sensible and non-breaking solution that I should have requested in the first place.

mattBrzezinski:
bors try

mattBrzezinski:
> Yes, that would be the sensible and non-breaking solution that I should have requested in the first place.
> I have modified the changes. What do you think?

LGTM! Just letting the CI run it!

jademackay:
Great!

mattBrzezinski:
Merging this in, ran `bors` and it succeeded. Idk why the check is having issues. I'll log one for it though.

PR #121, opened 2020-12-09
Cleanup the test CI buckets

mattBrzezinski:
For some reason buckets were not being cleaned up after the CI from the `S3Path` testing. I emptied and deleted all of these testing buckets.

These changes will now properly clean up after themselves. I'm not sure if we want to implement the `s3_nuke_bucket` as an exported function and give people the power explicitly.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -824,6 +824,21 @@ end
 s3_sign_url(a...;b...) = s3_sign_url(global_aws_config(), a...;b...)
 
 
+"""
+    s3_nuke_bucket(bucket_name)
+
+This function is NOT exported on purpose.
+It will delete all versions of objects in the given bucket and then the bucket itself.
+"""
+function s3_nuke_bucket(aws::AWSConfig, bucket_name)
+    for v in s3_list_versions(aws, bucket_name)
+        s3_delete(aws, bucket_name, v["Key"]; version = v["VersionId"])
+    end
+
+    s3_delete_bucket(aws, bucket_name)
+end
+
+
 include("s3path.jl")
 
 end #module AWSS3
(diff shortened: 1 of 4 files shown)


thomasgudjonwright:
Should `s3_nuke_bucket` be tested at all itself?

mattBrzezinski:
> Should `s3_nuke_bucket` be tested at all itself?

It's a test function, so I think it's fine. If we do choose to introduce this as a exported function, then yes.

test/awss3.jl
```diff
@@ -182,12 +181,7 @@ end
 end
 
 @testset "Empty and Delete Bucket" begin
-    for v in s3_list_versions(aws, bucket_name)
-        s3_delete(aws, bucket_name, v["Key"]; version = v["VersionId"])
-    end
-
-    s3_delete_bucket(aws, bucket_name)
-    @test !in(bucket_name, s3_list_buckets(aws))
```
mjram0s, line 190: Should you add this test back? Or was it testing wrong?

mattBrzezinski: This test is really just fluff, because the next test underneath this attempts to delete an already deleted bucket.

mjram0s (approved):

src/AWSS3.jl
```diff
@@ -824,6 +824,21 @@ end
 s3_sign_url(a...;b...) = s3_sign_url(global_aws_config(), a...;b...)
 
 
+"""
+    s3_nuke_bucket(bucket_name)
+
+This function is NOT exported on purpose.
```
mjram0s, line 830: Hm why do we not want to export this? Can explain more in the docstring.

mjram0s (approved):

src/AWSS3.jl
mattBrzezinski: Updated the docstr explaining. Just as a half-measure for people blowing up stuff by accident.

PR #118, opened 2020-12-07
multipart Base.write if large content

kolia:
avoiding

` HTTPClientError('An HTTP Client raised an unhandled exception: string longer than 2147483647 bytes')`

PR #128, opened 2021-01-01
write defaults to multipart for large files.

kolia:
To align with Base.write

(shoulda made this the default in #118 my bad)

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -349,7 +349,7 @@ function Base.write(fp::S3Path, content::Vector{UInt8}; part_size_mb=50, multipa
         return s3_put(fp.config, fp.bucket, fp.key, content)
     else
         io = IOBuffer(content)
-        return s3_multipart_upload(fp.config, fp.bucket, fp.key, io, part_size_mb=part_size_mb; other_kwargs...)
+        return s3_multipart_upload(fp.config, fp.bucket, fp.key, io, part_size_mb; other_kwargs...)
     end
 end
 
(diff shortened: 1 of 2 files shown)


kolia:
bump / any reason not to give this a bors try?

src/s3path.jl
```diff
@@ -342,14 +342,14 @@ Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key))
 
 Base.write(fp::S3Path, content::String; kwargs...) = Base.write(fp, Vector{UInt8}(content); kwargs...)
 
-function Base.write(fp::S3Path, content::Vector{UInt8}; part_size_mb=50, multipart::Bool=false, other_kwargs...)
+function Base.write(fp::S3Path, content::Vector{UInt8}; part_size_mb=50, multipart::Bool=true, other_kwargs...)
```
mattBrzezinski, line 345: Is there a reason to force the multi-part by default? I feel like there's probably more overhead to making that call vs. a single one if our content isn't too large.

mattBrzezinski:
> bump / any reason not to give this a bors try?

No reason other than being busy with other work.

mattBrzezinski:
bors try

kolia:
Thanks!

src/s3path.jl
ericphanson: it seems with the check a few lines later that even if `multipart` is true here, if `length(content) < MAX_HTTP_BYTES` holds then we still use just a `s3_put`. So my read of it is it only changes the default for large files, since small files use `s3_put` anyway.

ericphanson:
Let's see if this still works:

bors try

ericphanson (approved):

ericphanson:
This changes the default for `write(::S3Path, ...)` to use multipart writes for large files, and continue using single part writes for small files. I don't really think we've committed to *how* `write(::S3Path, ...)` is implemented; I think S3Paths are high enough level that this should be considered an implementation detail and thus non-breaking.

omus (approved):
LGTM. Will merge tomorrow morning

PR #129, opened 2021-01-11
add `join` keyword to `readdir`

ericphanson:
This keyword was added to Base in https://github.com/JuliaLang/julia/pull/33113, which made it into Julia 1.4.

I couldn't run the tests locally, but I tested with an example and it seemed to work. The new tests in `verify_files(path::AbstractPath)` need the Base method, so will fail on Julia 1 - 1.3. Should I put them behind a `VERSION` check, or just drop them?

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -336,7 +336,7 @@ function Base.readdir(fp::S3Path; join=false)
         end
 
         # Lexographically sort the results
-        return sort!(filter!(!isempty, unique!(names)))
+        return sort!(filter!(!isempty, names))
     else
         throw(ArgumentError("\"$fp\" is not a directory"))
     end
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -322,17 +322,21 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
     end
 end
 
-function Base.readdir(fp::S3Path)
+function Base.readdir(fp::S3Path; join=false)
     if isdir(fp)
         k = fp.key
         # Only list the files and "dirs" within this S3 "dir"
         objects = s3_list_objects(fp.config, fp.bucket, k; delimiter="")
 
         # Only list the basename and not the full key
-        basenames = unique!([s3_get_name(k, string(o["Key"])) for o in objects])
+        names = unique!([s3_get_name(k, string(o["Key"])) for o in objects])
+
+        if join
+            names = [ FilePathsBase.join(fp, name) for name in names]
+        end
 
         # Lexographically sort the results
-        return sort!(filter!(!isempty, basenames))
+        return sort!(filter!(!isempty, unique!(names)))
     else
         throw(ArgumentError("\"$fp\" is not a directory"))
     end
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -331,12 +331,14 @@ function Base.readdir(fp::S3Path; join=false)
         # Only list the basename and not the full key
         names = unique!([s3_get_name(k, string(o["Key"])) for o in objects])
 
+        # Lexographically sort the results
+        sort!(filter!(!isempty, names))
+        
         if join
             names = [ FilePathsBase.join(fp, name) for name in names]
         end
 
-        # Lexographically sort the results
-        return sort!(filter!(!isempty, names))
+        return names
     else
         throw(ArgumentError("\"$fp\" is not a directory"))
     end
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -322,23 +322,23 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
     end
 end
 
-function Base.readdir(fp::S3Path; join=false)
+function Base.readdir(fp::S3Path; join=false, sort=true)
     if isdir(fp)
         k = fp.key
         # Only list the files and "dirs" within this S3 "dir"
         objects = s3_list_objects(fp.config, fp.bucket, k; delimiter="")
 
         # Only list the basename and not the full key
-        names = unique!([s3_get_name(k, string(o["Key"])) for o in objects])
+        results = unique!([s3_get_name(k, string(o["Key"])) for o in objects])
 
-        # Lexographically sort the results
-        sort!(filter!(!isempty, names))
-        
-        if join
-            names = [ FilePathsBase.join(fp, name) for name in names]
-        end
+        # Filter out any empty object names which are valid in S3
+        filter!(!isempty, results)
 
-        return names
+        # Sort results if sort=true
+        sort && sort!(results)
+
+        # Return results, possibly joined with the root path if join=true
+        return join ? joinpath.(fp, results) : results
     else
         throw(ArgumentError("\"$fp\" is not a directory"))
     end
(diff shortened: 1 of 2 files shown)


mattBrzezinski:
bors try

mattBrzezinski:
> I couldn't run the tests locally, but I tested with an example and it seemed to work.

Any particular errors which popped up? I assume that it's probably a credential thing. At the very least you need credentials in ENV VARs / `~/.aws/config` / `~/.aws/credentials` files with full S3 perms.

> The new tests in `verify_files(path::AbstractPath)` need the Base method, so will fail on Julia 1 - 1.3. Should I put them behind a `VERSION` check, or just drop them?

I think keeping this behind a `VERSION` check would be best. Keeping packages on the same LTS as JuliaLang is the most easy way of handling things. Once 1.6 comes out as LTS, we can then dump the version check and set a new min ver.

ericphanson:
> Any particular errors which popped up? I assume that it's probably a credential thing. At the very least you need credentials in ENV VARs / ~/.aws/config / ~/.aws/credentials files with full S3 perms.

Ah, this would be it, I'm on a restricted role. I could set something up but for this PR hopefully CI will be enough. Would you mind re-running bors? I think I fixed the issues in my last two commits.

mattBrzezinski:
bors try

ericphanson:
sorry, looks like I had a copy-paste error there, try again?

mattBrzezinski:
bors try

ericphanson:
Bump / Is anything I need to do here? 🙂

mattBrzezinski:
@rofinn has more strong opinions on `S3Path` stuff than I do.
I think this is fine, maybe locking the `join=true` as a separate function behind a VERSION check would be my only comment.

src/s3path.jl
```diff
@@ -322,17 +322,23 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
     end
 end
 
-function Base.readdir(fp::S3Path)
+function Base.readdir(fp::S3Path; join=false)
```
rofinn, line 325: Maybe add `sort=true` while you're at it?

```
  │ Julia 1.4
  │
  │  The join and sort keyword arguments require at least Julia 1.4.
```

rofinn: We should also add this to FilePathsBase and test it in the test suite.

https://github.com/rofinn/FilePathsBase.jl/issues/110

src/s3path.jl
```diff
@@ -322,17 +322,23 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
 
         # Lexographically sort the results
-        return sort!(filter!(!isempty, basenames))
+        sort!(filter!(!isempty, names))
+        
+        if join
+            names = [ FilePathsBase.join(fp, name) for name in names]
+        end
+
+        return names
```
rofinn, line 341: This could probably be written more simply as:

```julia
# Only list the basename and not the full key
results = unique!([s3_get_name(k, string(o["Key"])) for o in objects])

# Filter out any empty object names which are valid in S3
filter!(!isempty, results)

# Sort results if sort=true
sort && sort!(results)

# Return results, possibly joined with the root path if join=true
return join ? joinpath.(fp, results) : results
```

rofinn (approved):
Seems fine to me. Just a couple minor suggestions, I've also made an issue to add this to FilePathsBase.jl as well.

rofinn:
@mattBrzezinski Do the tests not run anymore for this repo? I know Sam O'Connor ran the live tests in his personal account, but I thought we changed that when we started using bors?

mattBrzezinski:
> @mattBrzezinski Do the tests not run anymore for this repo? I know Sam O'Connor ran the live tests in his personal account, but I thought we changed that when we started using bors?

Yes they work w/ `bors` using the Invenia `Public CI` account. I assume you're referring to the original body. Eric had responded in this thread;

> Ah, this would be it, I'm on a restricted role. I could set something up but for this PR hopefully CI will be enough.

src/s3path.jl
ericphanson: looks good!

ericphanson:
thanks for the review! I've made those changes.

src/s3path.jl
```diff
@@ -322,17 +322,23 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
-        # Lexographically sort the results
-        return sort!(filter!(!isempty, basenames))
+        # Filter out any empty object names which are valid in S3
+        filter!(!isempty, results)
+
+        # Sort results if sort=true
+        sort && sort!(results)
+
+        # Return results, possibly joined with the root path if join=true
+        return join ? joinpath.(fp, results) : results
```
rofinn, line 341: Hmm, one thing I just realized about this is that it introduces a type instability cause `join` determines whether we return strings or paths.

ericphanson: Yeah, it does-- I've been using this and for me having the functionality has been well worth it, but it might depend on your use-case. Generally I find my code ends up like
```julia
for file in readdir(path; join=true)
    # some kind of filtering...
    endswith(string(file), ".arrow") || continue
    process_file!(results, file)
end
```
and the function-barrier (`process_file!`) means the instability is not an issue.

rofinn: I'm trying to remember, is there a reason you didn't want to use `readpath`? I'm wondering if for now I should just add those flags to both `readpath` and `readdir`, but just make the only difference that one return paths and one returns strings? In the future, I may make a breaking release which unifies the API and always returns paths for `readdir`.

ericphanson: Ah, is that what `readpath` does? I didn't know that function exists and it's not really documented:
```julia
julia> using FilePathsBase

help?> readpath
search: readpath

  readpath(fp::P) where {P <: AbstractPath} -> Vector{P}
```

Usually I stick to the Base API since I might have a mix of S3Paths and Base (string) paths.

(To be clear, `join` doesn't just return S3Path's vs strings, but rather it determines or not it's the full path, including the bucket etc, or just the filename).

rofinn: > (To be clear, join doesn't just return S3Path's vs strings, but rather it determines or not it's the full path, including the bucket etc, or just the filename).

Yeah, I was just wondering if it would have made more sense to do `joinpath.(string(fp), results)` in the `readdir` function and left returning paths to `readpath`?

rofinn: Okay. I've opted away from the type instability in FilePathsBase since I have less control over how it'll be used. It also means I can support that keyword without introducing a breaking change. I may change `readdir` to always return a path type if a path type is given in the future... in which case I'll plan to update `S3Path` as well.

https://github.com/rofinn/FilePathsBase.jl/pull/152

ericphanson: > I may change readdir to always return a path type if a path type is given

is there a way to do that while matching Base’s semantics? Like a RelativePath type or something?

rofinn: I've looked into the `RelativePath` idea, and it's very verbose. Without a proper traits based approach I don't think that's gonna happen anytime soon. Mostly likely I'd just wrap the resulting string in an appropriate path type and expect subtypes to decide what they want to do. I think there's also an issue where folks want `basename` to return a path as well.

PR #136, opened 2021-01-18
Tweaks to CI

omus:


omus:
bors try

fchorney:
At this point I think all the `1.3` stuff should be `1.5`? A little redundant since `1` is `1.5` right now, but will change to 1.6 soon enough right?

fchorney:
Might also want to implement these cache changes https://github.com/invenia/Hyperparameters.jl/pull/8

fchorney:
Also the slack notification if should be `always() && github.event_name == 'schedule'` instead of `github.event == 'schedule'`

omus:
> At this point I think all the `1.3` stuff should be `1.5`? A little redundant since `1` is `1.5` right now, but will change to 1.6 soon enough right?

You're not wrong. I'm planning on backporting this commit to AWSS3 0.6 so I'd rather keep 1.3 testing enabled for that.

.github/workflows/CI.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     name: Notify Slack Failure
     needs: test
     runs-on: ubuntu-latest
-    if: github.event == 'schedule'
+    if: always() && github.event == 'schedule'
```
fchorney, line 61:
```suggestion
    if: always() && github.event_name == 'schedule'
```

omus: That's what I get for being hasty

omus:
bors try

omus:
bors try

mattBrzezinski (approved):

mattBrzezinski:
Just needed the magic touch ;)

PR #140, opened 2021-01-30
support for `AbstractAWSConfig`

ExpandingMan:
This adds support for `AbstractAWSConfig` (see [here](https://github.com/JuliaCloud/AWS.jl/pull/274)) and adds some minimal documentation of `S3Path` to the README.  The changes here are minimal: `AWSConfig` arguments are generalized to `AbstractAWSConfig` and `S3Path` now accepts a single type parameter which is the type of the config.

Note that currently there is a limitation in `FilePathsBase` that prevents types with parameters from being used as paths.  That issue is fixed [here](https://github.com/rofinn/FilePathsBase.jl/pull/111).  `S3Path` will not work properly until that is merged and tagged (it's a very small change so I'm hoping that can happen quite soon).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "AWSS3"
 uuid = "1c724243-ef5b-51ab-93f4-b0a88ac62a95"
-version = "0.8.2"
+version = "0.8.3"
 
 [deps]
 AWS = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
--- a/Project.toml
+++ b/Project.toml
@@ -17,7 +17,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 XMLDict = "228000da-037f-5747-90a9-8195ccbf91a5"
 
 [compat]
-AWS = "1"
+AWS = "1.25"
 EzXML = "0.9, 1"
 FilePathsBase = "0.9"
 HTTP = "0.8, 0.9"
--- a/README.md
+++ b/README.md
@@ -19,3 +19,44 @@ s3_enable_versioning(aws, "my.bucket")
 s3_put(aws, "my.bucket", "key", "Hello!")
 println(s3_get(aws, "my.bucket", "key"))
 ```
+
+## `S3Path`
+This package exports the `S3Path` object.  This is an `AbstractPath` object as defined by
+[FilePathsBase.jl](https://github.com/rofinn/FilePathsBase.jl), allowing users to use
+Julia's `Base` [file system interface](https://docs.julialang.org/en/v1/base/file/) to
+obtain information from S3 buckets.  See the below example.
+```julia
+julia> using AWSS3, AWS, FilePathsBase;
+
+# global_aws_config() is also the default if no `config` argument is passed
+julia> p = S3Path("s3://bucket-name/dir1/", config=global_aws_config());
+
+julia> readdir(p)
+1-element Vector{SubString{String}}:
+ "demo.txt"
+
+julia> file = joinpath(p, "demo.txt")
+p"s3://bucket-name/dir1/demo.txt"
+
+julia> stat(file)
+Status(
+  device = 0,
+  inode = 0,
+  mode = -rw-rw-rw-,
+  nlink = 0,
+  uid = 1000 (username),
+  gid = 1000 (username),
+  rdev = 0,
+  size = 34 (34.0),
+  blksize = 4096 (4.0K),
+  blocks = 1,
+  mtime = 2021-01-30T18:53:02,
+  ctime = 2021-01-30T18:53:02,
+)
+
+julia> String(read(file))  # fetch the file into memory
+"this is a file for testing S3Path\n"
+
+julia> rm(file)  # delete the file
+UInt8[]
+```
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ This package exports the `S3Path` object.  This is an `AbstractPath` object as d
 [FilePathsBase.jl](https://github.com/rofinn/FilePathsBase.jl), allowing users to use
 Julia's `Base` [file system interface](https://docs.julialang.org/en/v1/base/file/) to
 obtain information from S3 buckets.  See the below example.
+
 ```julia
 julia> using AWSS3, AWS, FilePathsBase;
 
--- a/README.md
+++ b/README.md
@@ -21,10 +21,11 @@ println(s3_get(aws, "my.bucket", "key"))
 ```
 
 ## `S3Path`
-This package exports the `S3Path` object.  This is an `AbstractPath` object as defined by
-[FilePathsBase.jl](https://github.com/rofinn/FilePathsBase.jl), allowing users to use
+This package exports the `S3Path` object. 
+This is an `AbstractPath` object as defined by [FilePathsBase.jl](https://github.com/rofinn/FilePathsBase.jl), allowing users to use
 Julia's `Base` [file system interface](https://docs.julialang.org/en/v1/base/file/) to
-obtain information from S3 buckets.  See the below example.
+obtain information from S3 buckets.
+See the below example.
 
 ```julia
 julia> using AWSS3, AWS, FilePathsBase;
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -7,9 +7,6 @@
 #==============================================================================#
 
 
-__precompile__()
-
-
 module AWSS3
 
 export S3Path, s3_arn, s3_put, s3_get, s3_get_file, s3_exists, s3_delete, s3_copy,
@@ -54,7 +51,7 @@ s3_arn(bucket, path) = s3_arn("$bucket/$path")
 
 
 """
-    s3_get([::AWSConfig], bucket, path; <keyword arguments>)
+    s3_get([::AbstractAWSConfig], bucket, path; <keyword arguments>)
 
 [Get Object](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html)
 from `path` in `bucket`.
@@ -71,7 +68,7 @@ For example, to get a `range` of bytes instead of the whole object, do:
 
 `s3_get(aws, bucket, path; headers=Dict{String,String}("Range" => "bytes=\$(first(range)-1)-\$(last(range)-1)"))`
 """
-function s3_get(aws::AWSConfig, bucket, path; version="", retry=true, raw=false, headers=Dict{String, Any}(), kwargs...)
+function s3_get(aws::AbstractAWSConfig, bucket, path; version="", retry=true, raw=false, headers=Dict{String, Any}(), kwargs...)
     @repeat 4 try
         args = Dict{String, Any}(
             "return_raw" => raw
@@ -95,11 +92,11 @@ s3_get(a...; b...) = s3_get(global_aws_config(), a...; b...)
 
 
 """
-    s3_get_file([::AWSConfig], bucket, path, filename; [version=], kwargs...)
+    s3_get_file([::AbstractAWSConfig], bucket, path, filename; [version=], kwargs...)
 
 Like `s3_get` but streams result directly to `filename`.
 """
-function s3_get_file(aws::AWSConfig, bucket, path, filename; version="", kwargs...)
+function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version="", kwargs...)
     stream = S3.get_object(bucket, path, Dict("return_stream"=>true); aws_config=aws, kwargs...)
 
     open(filename, "w") do file
@@ -112,7 +109,7 @@ end
 s3_get_file(a...; b...) = s3_get_file(global_aws_config(), a...; b...)
 
 
-function s3_get_file(aws::AWSConfig, buckets::Vector, path, filename; version="", kwargs...)
+function s3_get_file(aws::AbstractAWSConfig, buckets::Vector, path, filename; version="", kwargs...)
     i = start(buckets)
 
     @repeat length(buckets) try
@@ -125,13 +122,13 @@ end
 
 
 """
-   s3_get_meta([::AWSConfig], bucket, path; [version=], kwargs...)
+   s3_get_meta([::AbstractAWSConfig], bucket, path; [version=], kwargs...)
 
 [HEAD Object](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectHEAD.html)
 
 Retrieves metadata from an object without returning the object itself.
 """
-function s3_get_meta(aws::AWSConfig, bucket, path; version="", kwargs...)
+function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version="", kwargs...)
     if isempty(version)
         S3.head_object(bucket, path; aws_config=aws, kwargs...)
     else
@@ -143,11 +140,11 @@ s3_get_meta(a...; b...) = s3_get_meta(global_aws_config(), a...; b...)
 
 
 """
-    s3_exists([::AWSConfig], bucket, path [version=], kwargs...)
+    s3_exists([::AbstractAWSConfig], bucket, path [version=], kwargs...)
 
 Is there an object in `bucket` at `path`?
 """
-function s3_exists(aws::AWSConfig, bucket, path; version="", kwargs...)
+function s3_exists(aws::AbstractAWSConfig, bucket, path; version="", kwargs...)
     @repeat 2 try
         s3_get_meta(aws, bucket, path; version=version, kwargs...)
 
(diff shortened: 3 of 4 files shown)


ExpandingMan:
Note that the requisite PR to `FilePathsBase` has already been merged and tagged, so there should be no issue there.

ExpandingMan:
AWS has now been tagged, so this should be fully working now.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "AWSS3"
 uuid = "1c724243-ef5b-51ab-93f4-b0a88ac62a95"
-version = "0.8.2"
+version = "0.8.3"
```
mattBrzezinski, line 3: GH is not allowing me to comment further down in the file.
However, do we not need to bump:

```toml
[compat]
AWS = "1.25"
```

README.md
```diff
@@ -19,3 +19,44 @@ s3_enable_versioning(aws, "my.bucket")
 s3_put(aws, "my.bucket", "key", "Hello!")
 println(s3_get(aws, "my.bucket", "key"))
 ```
+
+## `S3Path`
+This package exports the `S3Path` object.  This is an `AbstractPath` object as defined by
+[FilePathsBase.jl](https://github.com/rofinn/FilePathsBase.jl), allowing users to use
+Julia's `Base` [file system interface](https://docs.julialang.org/en/v1/base/file/) to
+obtain information from S3 buckets.  See the below example.
+```julia
```
mattBrzezinski, line 28:
```suggestion

```julia
```

src/AWSS3.jl
```diff
@@ -7,9 +7,6 @@
 #==============================================================================#
 
 
-__precompile__()
```
mattBrzezinski, line 10: Any reason to take this out?

Project.toml
ExpandingMan: done

README.md
ExpandingMan: done

src/AWSS3.jl
ExpandingMan: It has been unnecessary (precompile true by default, same as `__precompile__()` since Julia version 0.7.  Since the compat bound is 1 anyway, it is not necessary on any version of Julia that this package will work on.  Seemed superfluous.

mattBrzezinski: Ah interesting, didn't know that!

mattBrzezinski:
bors try

ExpandingMan:
Looks like I broke some of the `S3Path` constructors.  Will fix.

ExpandingMan:
We have a bigger problem here, though apparently it's not a new one.

So, if you do `FilePathsBase.register`, this does a `pushfirst!` onto a global variable.  Therefore, any time you do, e.g. `p"/usr/bin/exe"`, the very first thing that will happen is that the program tries to talk to AWS.  This is really bad, for all sorts of reasons, not the least of which is that if you happen to not be on AWS (as is my case here where I'm testing with min.io) this will basically hang for a super long time trying and failing to talk to an EC2 server.

I strongly suggest that this package should *NOT* call `register` by default.  Instead, we should include in the docs that users have the option to do this, but make sure they understand the consequences.

Do you agree with this?  If so, I will make the change in this PR.  Otherwise, I will try to get tests passing, and we can figure it out after this gets merged.

ExpandingMan:
Alright, so I think I fixed the error that we originally saw on bors.  I'm a little confused by it, becuase it seems like it must have been an error all along, but regardless, should now be fixed.

Additionally, I've implemented a little hack to make sure that FilePathsBase doesn't inadvertently cause calls to AWS.  The `tryparse` method for `S3Path` now accepts `nothing` as an argument for `config` and this is the new default.  If it gets `nothing`, it will first check if the string is an S3 string before attempting to construct the config.  This prevents code in FilePathsBase from trying to talk to AWS to get to an AWS config unless the user has a `s3://` path string.

If it were up to me, I would still remove the `FilePathsBase.register` call from the `__init__()` of this package.  It just seems a little too weird that `p"s3://bucket/path` can start talking to AWS.  If we removed it users could still call it, but there would be fewer nasty surprises.

In any case, that would be a breaking change, and what I have here now should be non-breaking, and also eliminate any crazy behavior.  Let me know what you think about `register`.  If you want to change it, I can change it now, or we can merge this and tag it and change it afterwards as a breaking change.

I don't know if bors will work for me but I'll give it a shot so we can see results.

ExpandingMan:
bors try

ExpandingMan:
Ugh, that was because I used `isnothing` which is not available on whatever old version of Julia this was tested on.  I changed it, try again when ready.

mattBrzezinski:
bors try

ExpandingMan:
Man, doing this without being able to run unit tests locally sure is painful.

Anyway, that was because the `Type` argument to `tryparse` was inappropriately strict.  Should be fixed, try again when ready

mattBrzezinski:
bors try

ExpandingMan:
Ok, so apparently FilePathsBase utilizes a couple of different constructors called directly with the type (e.g. `typeof(arg)()`).  I have added the appropriate methods.  The constructor methods available may now seem a little messy, but I have tried to make it completely non-breaking while also always handling the conversion for the first argument.

Once more unto the breach?

mattBrzezinski:
bors try

src/s3path.jl
```diff
@@ -71,15 +85,18 @@ function S3Path(
...
-function S3Path(str::AbstractString; config::AWSConfig=global_aws_config())
+function S3Path(str::AbstractString; config::AbstractAWSConfig=global_aws_config())
     result = tryparse(S3Path, str; config=config)
     result !== nothing || throw(ArgumentError("Invalid s3 path string: $str"))
     return result
 end
 
-function Base.tryparse(::Type{S3Path}, str::AbstractString; config::AWSConfig=global_aws_config())
+# if config=nothing, will not try to talk to AWS until after string is confirmed to be an s3 path
+function Base.tryparse(::Type{<:S3Path}, str::AbstractString; config::Union{Nothing,AbstractAWSConfig}=nothing)
```
mattBrzezinski, line 95: Can this not be?

> ; config::AbstractAWSConfig=global_aws_config())

Then remove the `(config === nothing)`

src/s3path.jl
```diff
@@ -1,14 +1,26 @@
...
     segments::Tuple{Vararg{String}}
     root::String
     drive::String
     isdirectory::Bool
-    config::AWSConfig
+    config::A
+
+    # this method ensures that there is proper conversion to Tuple{Vararg{String}} for first arg,
+    # which can be a little tempremental if you don't explicitly handle it
+    function S3Path{A}(segments, root::AbstractString, drive::AbstractString, isdirectory::Bool,
```
mattBrzezinski, line 10: Is there some more context to this to adding this in? Just not following how the change from `AWSConfig -> AbstractAWSConfig` necessitates this

src/s3path.jl
```diff
@@ -1,14 +1,26 @@
...
+                       config::A) where {A<:AbstractAWSConfig}
```
mattBrzezinski, line 11:
```suggestion
    function S3Path{A}(
        segments,
        root::AbstractString,
        drive::AbstractString,
        isdirectory::Bool,
        config::A
    ) where {A<:AbstractAWSConfig}
```

mattBrzezinski: ....GH suggestions just don't seem to be happy. Breaking this into the 2nd multi-line if needed fits better with the style guide

src/s3path.jl
ExpandingMan: No, that's the whole point.

What was happening here was that any time you try to parse a path, that is *ANY* path with `p"/a/b/c"`, because of the way `FilePathsBase.register` works the *very first thing* that would happen (if you don't have AWS environment variables) is that it starts trying to talk to the AWS API to try to get EC2 credentials.

So for example, I was testing this on one of my machines with min.io.  I have no AWS environment variables and am not an AWS's environment.  So, if I do `p"/etc/fstab"` (yes, even a filesystem path) it hangs for a long time and then ultimately errors because it called `global_aws_config()`, did not have local environment variables, and could not talk to the AWS end point where it would get EC2 credentials.  Arguably, this would be bad even on AWS because an unexpectedly expensive network call is taking place even when you are doing something as innocent as declaring a file system path.

The work around you see here partially solves the problem by ensuring that `global_aws_config()` is *NOT* called unless the path it is trying to parse really is an S3 path.

Again, I personally find even this a little dubious, and if it were up to me I'd remove `FilePathsBase.register` from the initialization of AWSS3 (users who wanted it would still have the option of adding it themselves).  However, the changes I have made above are non-breaking, so there's no need to decide whether you agree with me on that right now.

src/s3path.jl
ExpandingMan: It's rather subtle.  It turns out that `FilePathsBase` calls type constructors directly somewhere.  For example, if you have a function
```julia
f(::Type{T}) where {T} = T()
```
if `T` is `S3Path{AWSConfig}`, this will fail unless the above method is defined, because this is a *different method* from `S3Path()` (because `S3Path{AWSConfig}` and `S3Path` are different types, the latter is an implicit abstract type).

I concede that the constructors I've wound up with here are a little messy, but I have included them because they should
- Ensure there are no breaking changes for the `S3Path` constructor, i.e. the user interface of AWSS3.jl has no breaking changes from this PR.
- Ensure that `FilePathsBase` has the methods it needs for the calls it makes.
- Ensure that the first argument of `S3Path` always gets converted to the correct type, which is a little subtle (due to it being a bit of a weird type) and seemed to be error prone before this PR.

src/s3path.jl
ExpandingMan: That seems rather excessively verbose, but ok, committing.

src/s3path.jl
mattBrzezinski: Ah okay, thank you for clarifying! This solution definitely makes sense to go with, and also should lead to some expanse in our unit testing to run a some suite w/o credentials so we can keep track of these things.

ExpandingMan: Great, thanks.  Yeah, some more tests are probably needed.  I'm not going to offer to add any since it's so hard for me to test this package.

Btw, after this is merged, I'm going to publish Minio.jl, which is just a very small package that makes it convenient to declare a config for minio, and also makes it convenient to fetch and run the minio binary as a server.  Part of why I set that up is to make testing easier, and testing this package could be a possible application of this.

mattBrzezinski: Logged an issue for it:

- https://github.com/JuliaCloud/AWSS3.jl/issues/142

I'll get around to it at some point

ExpandingMan:
Looks like we are resolved then.  I suggest you run a bors one more time just to sanity check that one syntax change.

After merging, a tag would be hugely appreciated, because like I said, I have a small package I'm waiting on this to register.

Thanks!

src/s3path.jl
```diff
@@ -1,14 +1,31 @@
...
+    # this method ensures that there is proper conversion to Tuple{Vararg{String}} for first arg,
+    # which can be a little tempremental if you don't explicitly handle it
+    function S3Path{A}(
+        segments,
+        root::AbstractString,
+        drive::AbstractString,
+        isdirectory::Bool,
+        config::A
+    ) where {A<:AbstractAWSConfig}
+        new{typeof(config)}(tuple((String(s) for s ∈ segments)...), root, drive, isdirectory, config)
```
mattBrzezinski, line 17: Just want to cc @rofinn here since he is the mastermind behind this pathing.
Not sure if opinions are had here or not.

mattBrzezinski, line 17:
```suggestion
        new{typeof(config)}(tuple((String(s) for s in segments)...), root, drive, isdirectory, config)
```

mattBrzezinski (approved):

src/s3path.jl
ExpandingMan: 😞

ExpandingMan: The issue is that `Tuple{Vararg{String}}` is much stricter than other types (e.g. `Vector`) with conversions, so one has to be careful with e.g. `Tuple{Vararg{SubString}}`.  I considered changing the type of the field, since it's not even a concrete type, but I noticed that this is the same field type as those in `FilePathsBase` so I figured it was best to leave it alone and just make sure it wasn't going to crash.

src/s3path.jl
```diff
@@ -1,14 +1,31 @@
...
+struct S3Path{A<:AbstractAWSConfig} <: AbstractPath
     segments::Tuple{Vararg{String}}
     root::String
     drive::String
     isdirectory::Bool
-    config::AWSConfig
+    config::A
+
+    # this method ensures that there is proper conversion to Tuple{Vararg{String}} for first arg,
+    # which can be a little tempremental if you don't explicitly handle it
```
rofinn, line 9:
```suggestion
    # which can be a little temperamental if you don't explicitly handle it
```

rofinn: Can you clarify what the issue was?

src/s3path.jl
```diff
@@ -1,14 +1,31 @@
...
+    function S3Path{A}(
+        segments,
+        root::AbstractString,
+        drive::AbstractString,
+        isdirectory::Bool,
+        config::A
+    ) where {A<:AbstractAWSConfig}
+        new{typeof(config)}(tuple((String(s) for s in segments)...), root, drive, isdirectory, config)
```
rofinn, line 17: I don't love that we need to do this iteration, conversion and splatting every time we construct the path type? Can we have an outer constructor that addresses the specific case where this is necessary (e.g., segments aren't `Tuple{Vararg{String}}`?

rofinn (approved):
Seems mostly fine to me, but I don't like addressing the segments special case in an inner constructor. I think an appropriate dispatch for the outer constructors would suffice.

src/s3path.jl
ExpandingMan: I was in the process of making a simple MWE to show you why it so easily fails, but to my surprise, in my simple example I was not having the issues I had with this that prompted me to make this change.  I'll experiment with this a bit and see if I can come up with a better solution.

rofinn: FWIW, if you find the failure case it might be worth adding a test here and opening an issue against FilePathsBase.jl? Even if we can't fix the problem there generically, we could at least fix the system paths and include it in the test suite.

ExpandingMan:
Sorry about that, turns out I had misdiagnosed the problem I was having.  The problem was *not* that Julia was having trouble converting the type of the first argument, the problem was that I was not properly directing the call to the method that converts.

I have removed the inner constructor, and I have added a simple method (without any splatting) to ensure that the constructor still performs proper conversions even if no type parameter is provided.

Note that we seem to still need the `S3Path{T}() where {T}` constructor, as FilePathsBase likes to call constructors like `typeof(arg)()` and I wanted to keep `S3Path()` so as not to have any breaking changes (and might as well have it, right?).

Anyway, should be good now, please trigger bors when ready.

rofinn:
> Note that we seem to still need the S3Path{T}() where {T} constructor, as FilePathsBase likes to call constructors like typeof(arg)() and I wanted to keep S3Path() so as not to have any breaking changes (and might as well have it, right?).

Cool, yeah, that makes sense to me :)

ExpandingMan:
> FWIW, if you can think of a better way to handle the registration/Path(...)/p_str logic feel free to open an issue. That stuff was very much a means to an end...

I think it's fine like it is, I just think packages (particularly this one) should be more conservative about how they call `FilePathsBase.register`.  Unfortunately using a `@p_str` macro severely restricts how you can pass it context, so I don't really see a better way of doing it than what you did.  It's a nice option to have if you really want it, but I'd think in many situations it would make more sense to just call the constructor.  I suppose you could do something clever in the macro like sniff out the environment, but that seems a little too magical, probably better just to give users more control.

Btw, if you guys are willing to remove `__init__() = FilePathsBase.register(S3Path)` in this PR and just suggest it in documentation, let me know.  In my mind there is no question that it should go, but it would be a very breaking change (anyone who made extensive use of `p"s3://a/b/c` would be in fur a rude update), and I kind of just want to get this merged, so I wasn't going to press the matter.

rofinn:
> anyone who made extensive use of p"s3://a/b/c would be in fur a rude update

Yeah, we use it pretty extensively and simply rely config files and environment variables to control the default behaviour.

mattBrzezinski:
bors try

mattBrzezinski:
Going over the new commits in the morning, and then I'll merge and tag!

mattBrzezinski:
Ran into an AWS timing issue, I'll retry.

ExpandingMan:
Thanks!  Just a gentle reminder to please not forget about a tag.  There should be no breaking changes so 0.8.3 should work.

ExpandingMan:
Oops, didn't notice that it was already on its way.  Thanks again!

rofinn:
Since this PR technically fixes a bug, would it make sense to backport at least the `tryparse` fix to 0.7?

mattBrzezinski:
> Since this PR technically fixes a bug, would it make sense to backport at least the `tryparse` fix to 0.7?

Created a PR for this here, https://github.com/JuliaCloud/AWSS3.jl/pull/144/files, not sure the proper way to backport though.

rofinn:
I think we typically just make a branch for the release (`0.7.7`)and make the tag off that. You should be able to run the julia registrator off that branch/commit though.

mattBrzezinski:
> I think we typically just make a branch for the release (`0.7.7`)and make the tag off that. You should be able to run the julia registrator off that branch/commit though.

Hmmm interesting, we would then need that branch to continuing living forward then outside of `master` I guess?

rofinn:
Not necessary. You just need the branch to exist long enough to tag the release. The commit will still exist if it isn't on a branch and folks can always make subsequent branches off of your tagged release.

oxinabox:
In retrospect making the config type a type parameter was not a great change.
It was breaking, and didn't give any real gain.

Making it a type parameter, rather than an abstract field avoids dynamic dispatch.
But dynamic dispatch costs 20-200ns.
In contrast, this is doing network IO, so it already takes 20-200us.

Issue #155, opened 2021-04-25
support `version` argument for S3Path


jrevels:
It'd be useful to refer to specific object versions when constructing and using `S3Path`s, like you currently can when using e.g. `s3_get` directly.

ericphanson:
Closed by #161 

(release isn't tagged yet but we will tag shortly with #182 that adds more support)

PR #161, opened 2021-06-03
Support `version` argument for S3Path

hannahilea:
Implements #155.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -36,6 +36,7 @@ using UUIDs
 @service S3
 
 const SSDict = Dict{String,String}
+const S3PathVersion = Union{String,Nothing}
 
 __init__() = FilePathsBase.register(S3Path)
 
@@ -74,7 +75,7 @@ As an example of how to set custom HTTP headers, the below is equivalent to
 `s3_get(aws, bucket, path; headers=Dict{String,String}("Range" => "bytes=\$(first(range)-1)-\$(last(range)-1)"))`
 """
 function s3_get(
-    aws::AbstractAWSConfig, bucket, path; version="", retry::Bool=true,
+    aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing, retry::Bool=true,
     byte_range::Union{Nothing,AbstractVector}=nothing, raw::Bool=false,
     headers::AbstractDict{<:AbstractString,<:Any}=Dict{String, Any}(),
     return_stream::Bool=false, kwargs...
@@ -85,7 +86,7 @@ function s3_get(
             "return_stream" => return_stream,
         )
 
-        if !isempty(version)
+        if !isnothing(version)
             args["versionId"] = version
         end
 
@@ -115,7 +116,7 @@ s3_get(a...; b...) = s3_get(global_aws_config(), a...; b...)
 Like `s3_get` but streams result directly to `filename`.  Keyword arguments accept are
 the same as those for `s3_get`.
 """
-function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version="", kwargs...)
+function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version::S3PathVersion=nothing, kwargs...)
     stream = s3_get(aws, bucket, path; return_stream=true, kwargs...)
 
     open(filename, "w") do file
@@ -128,7 +129,8 @@ end
 s3_get_file(a...; b...) = s3_get_file(global_aws_config(), a...; b...)
 
 
-function s3_get_file(aws::AbstractAWSConfig, buckets::Vector, path, filename; version="", kwargs...)
+function s3_get_file(aws::AbstractAWSConfig, buckets::Vector, path, filename;
+                     version::S3PathVersion=nothing, kwargs...)
     i = start(buckets)
 
     @repeat length(buckets) try
@@ -147,8 +149,9 @@ end
 
 Retrieves metadata from an object without returning the object itself.
 """
-function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version="", kwargs...)
-    if isempty(version)
+function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
+                     kwargs...)
+    if isnothing(version)
         S3.head_object(bucket, path; aws_config=aws, kwargs...)
     else
         S3.head_object(bucket, path, Dict("versionId"=>version); aws_config=aws, kwargs...)
@@ -163,7 +166,7 @@ s3_get_meta(a...; b...) = s3_get_meta(global_aws_config(), a...; b...)
 
 Is there an object in `bucket` at `path`?
 """
-function s3_exists(aws::AbstractAWSConfig, bucket, path; version="", kwargs...)
+function s3_exists(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing, kwargs...)
     @repeat 2 try
         s3_get_meta(aws, bucket, path; version=version, kwargs...)
 
@@ -187,8 +190,9 @@ s3_exists(a...; b...) = s3_exists(global_aws_config(), a...; b...)
 
 [DELETE Object](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectDELETE.html)
 """
-function s3_delete(aws::AbstractAWSConfig, bucket, path; version="", kwargs...)
-    if isempty(version)
+function s3_delete(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
+                   kwargs...)
+    if isnothing(version)
         S3.delete_object(bucket, path; aws_config=aws, kwargs...)
     else
         S3.delete_object(bucket, path, Dict("versionId"=>version); aws_config=aws, kwargs...)
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -36,7 +36,7 @@ using UUIDs
 @service S3
 
 const SSDict = Dict{String,String}
-const S3PathVersion = Union{AbstractString,Nothing}
+const AbstractS3Version = Union{AbstractString,Nothing}
 
 __init__() = FilePathsBase.register(S3Path)
 
@@ -75,7 +75,7 @@ As an example of how to set custom HTTP headers, the below is equivalent to
 `s3_get(aws, bucket, path; headers=Dict{String,String}("Range" => "bytes=\$(first(range)-1)-\$(last(range)-1)"))`
 """
 function s3_get(
-    aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing, retry::Bool=true,
+    aws::AbstractAWSConfig, bucket, path; version::AbstractS3Version=nothing, retry::Bool=true,
     byte_range::Union{Nothing,AbstractVector}=nothing, raw::Bool=false,
     headers::AbstractDict{<:AbstractString,<:Any}=Dict{String, Any}(),
     return_stream::Bool=false, kwargs...
@@ -115,7 +115,7 @@ s3_get(a...; b...) = s3_get(global_aws_config(), a...; b...)
 Like `s3_get` but streams result directly to `filename`.  Keyword arguments accept are
 the same as those for `s3_get`.
 """
-function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version::S3PathVersion=nothing, kwargs...)
+function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version::AbstractS3Version=nothing, kwargs...)
     stream = s3_get(aws, bucket, path; version=version, return_stream=true, kwargs...)
 
     open(filename, "w") do file
@@ -129,7 +129,7 @@ s3_get_file(a...; b...) = s3_get_file(global_aws_config(), a...; b...)
 
 
 function s3_get_file(aws::AbstractAWSConfig, buckets::Vector, path, filename;
-                     version::S3PathVersion=nothing, kwargs...)
+                     version::AbstractS3Version=nothing, kwargs...)
     i = start(buckets)
 
     @repeat length(buckets) try
@@ -148,7 +148,7 @@ end
 
 Retrieves metadata from an object without returning the object itself.
 """
-function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
+function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version::AbstractS3Version=nothing,
                      kwargs...)
     if version === nothing || isempty(version)
         S3.head_object(bucket, path; aws_config=aws, kwargs...)
@@ -165,7 +165,7 @@ s3_get_meta(a...; b...) = s3_get_meta(global_aws_config(), a...; b...)
 
 Is there an object in `bucket` at `path`?
 """
-function s3_exists(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing, kwargs...)
+function s3_exists(aws::AbstractAWSConfig, bucket, path; version::AbstractS3Version=nothing, kwargs...)
     @repeat 2 try
         s3_get_meta(aws, bucket, path; version=version, kwargs...)
 
@@ -189,7 +189,7 @@ s3_exists(a...; b...) = s3_exists(global_aws_config(), a...; b...)
 
 [DELETE Object](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectDELETE.html)
 """
-function s3_delete(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
+function s3_delete(aws::AbstractAWSConfig, bucket, path; version::AbstractS3Version=nothing,
                    kwargs...)
     if version === nothing || isempty(version)
         S3.delete_object(bucket, path; aws_config=aws, kwargs...)
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -117,7 +117,7 @@ Like `s3_get` but streams result directly to `filename`.  Keyword arguments acce
 the same as those for `s3_get`.
 """
 function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version::S3PathVersion=nothing, kwargs...)
-    stream = s3_get(aws, bucket, path; return_stream=true, kwargs...)
+    stream = s3_get(aws, bucket, path; version=version, return_stream=true, kwargs...)
 
     open(filename, "w") do file
         while !eof(stream)
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -86,7 +86,7 @@ function s3_get(
             "return_stream" => return_stream,
         )
 
-        if !isnothing(version)
+        if !isnothing(version) || !isempty(version)
             args["versionId"] = version
         end
 
@@ -151,7 +151,7 @@ Retrieves metadata from an object without returning the object itself.
 """
 function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
                      kwargs...)
-    if isnothing(version)
+    if isnothing(version) || isempty(version)
         S3.head_object(bucket, path; aws_config=aws, kwargs...)
     else
         S3.head_object(bucket, path, Dict("versionId"=>version); aws_config=aws, kwargs...)
@@ -192,7 +192,7 @@ s3_exists(a...; b...) = s3_exists(global_aws_config(), a...; b...)
 """
 function s3_delete(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
                    kwargs...)
-    if isnothing(version)
+    if isnothing(version) || isempty(version)
         S3.delete_object(bucket, path; aws_config=aws, kwargs...)
     else
         S3.delete_object(bucket, path, Dict("versionId"=>version); aws_config=aws, kwargs...)
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -85,7 +85,7 @@ function s3_get(
             "return_raw" => raw,
             "return_stream" => return_stream,
         )
-        if !isnothing(version) && !isempty(version)
+        if version !== nothing && !isempty(version)
             args["versionId"] = version
         end
 
@@ -150,7 +150,7 @@ Retrieves metadata from an object without returning the object itself.
 """
 function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
                      kwargs...)
-    if isnothing(version) || isempty(version)
+    if version === nothing || isempty(version)
         S3.head_object(bucket, path; aws_config=aws, kwargs...)
     else
         S3.head_object(bucket, path, Dict("versionId"=>version); aws_config=aws, kwargs...)
@@ -191,7 +191,7 @@ s3_exists(a...; b...) = s3_exists(global_aws_config(), a...; b...)
 """
 function s3_delete(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
                    kwargs...)
-    if isnothing(version) || isempty(version)
+    if version === nothing || isempty(version)
         S3.delete_object(bucket, path; aws_config=aws, kwargs...)
     else
         S3.delete_object(bucket, path, Dict("versionId"=>version); aws_config=aws, kwargs...)
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -85,8 +85,7 @@ function s3_get(
             "return_raw" => raw,
             "return_stream" => return_stream,
         )
-
-        if !isnothing(version) || !isempty(version)
+        if !isnothing(version) && !isempty(version)
             args["versionId"] = version
         end
 
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -36,7 +36,7 @@ using UUIDs
 @service S3
 
 const SSDict = Dict{String,String}
-const S3PathVersion = Union{String,Nothing}
+const S3PathVersion = Union{AbstractString,Nothing}
 
 __init__() = FilePathsBase.register(S3Path)
 
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -9,7 +9,7 @@ end
 
 # constructor that converts but does not require type parameter
 function S3Path(segments, root::AbstractString, drive::AbstractString, isdirectory::Bool,
-                config::AbstractAWSConfig, version::S3PathVersion=nothing)
+                config::AbstractAWSConfig, version::AbstractS3Version=nothing)
     S3Path{typeof(config)}(segments, root, drive, isdirectory, config, version)
 end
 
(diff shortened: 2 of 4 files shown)


src/s3path.jl
```diff
@@ -355,13 +363,14 @@ function Base.readdir(fp::S3Path; join=false, sort=true)
     end
 end
 
-Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key))
+Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key; version=fp.version))
```
omus, line 366: Hmmm, I'm starting to wonder if `version` should be part of an `S3Path` or just an argument of a function call. I'll need to think about this.

omus: This [particular line](https://github.com/JuliaCloud/AWSS3.jl/pull/161/files#diff-31b593db3c82f33730f4af58846ed59213199a8dd973bddc4f70f1e1721d6719R373) is making me question this design. As the version information is only useful for reading probably we should limit its use to just those kinds functions. I'm unsure on if that will result in higher-level functions not being able to support versioning then

test/s3path.jl
```diff
@@ -334,4 +335,33 @@ end
     end
 end
 
+@testset "S3Path versioning" begin
+    key_version_file = "test_versions"
+    s3_put(bucket_name, key_version_file, "data.v1")
+    s3_put(bucket_name, key_version_file, "data.v2")
+
+    versions = s3_list_versions(aws, bucket_name, key_version_file)
+    @test length(versions) == 2 # versions is ordered newest to oldest---`first(versions)` is the latest version
```
omus, line 344: Probably more intuitive to do:
```suggestion
    # `s3_list_versions` returns versions in the order newest to oldest
    versions = reverse!(s3_list_versions(aws, bucket_name, key_version_file))
    @test length(versions) == 2 
```
This will require additional changes below.

src/s3path.jl
mattBrzezinski: Having it as a optional argument feels right to me. But thinking about it a bit, when retrieving a specific version of an object it's part of the URL / path.

> https://{my-bucket}.s3.us-east-1.amazonaws.com/foobar.txt?versionId=BaCWQOBnUvNOjOBm3zY1aTVM0pD8yfdn

hannahilea: >This particular line

Yeah---I actually went back and forth on whether the correct behavior there would be to throw an error (or a warning) or just to silently write a new version. I don't have a strong opinion about the correct thing to do.

> As the version information is only useful for reading

Also useful for deleting a specific version! (Although admittedly that is not a use case I've run into much.)

hannahilea: > when retrieving a specific version of an object it's part of the URL / path

My use case (and that of others on my team) is such that a large number of the S3 objects that we work with (and there are many!!) _must_ be versioned---the version is a critical part of the object path---which is why I chose this particular implementation.

Until now, I've been using a wonky workaround to call `s3_get` directly for each of these items, e.g.,
```
const dataset_s3path = S3Path("s3://my_bucket/my_key")
const dataset_version = "aweofijwoefjieowjf"

my_dataset = s3_get(dataset_s3path.bucket, dataset_s3path.key; version=dataset_version)
```
This proposed implementation lets us pass around a single S3Path without needing to separately track/pull in a version:
```
const dataset_s3path = S3Path("s3://my_bucket/my_key"; version="aweofijwoefjieowjf")
...
my_dataset = read(dataset_s3path)
```

If `version` is an argument to `read` functions rather than a property of `S3Path` itself, we'll continue to need to separately track versions (and think carefully at all `read` callsites about whether or not to pull them in) for any given file:

```
const dataset_s3path = S3Path("s3://my_bucket/my_key")
const dataset_version = "aweofijwoefjieowjf"
...

my_dataset = read(dataset_s3path; version=dataset_version)
```

...this example is kind of goofy, and makes it seem like I'm averse to keeping track of one extra variable---in reality, we're often keeping track of many paths at the same time, some versioned and some not, which makes the process of additionally tracking versions fairly error-prone, especially when the version truly is a property of the S3 object.

mattBrzezinski: I haven't spent too much time thinking about this, I will in the next few days though. I think it does seem like a property of an `S3Path`, and would simplify things in your workflow.

I imagine that currently when working with multiple datasets it just ends up being something like:

```julia
const DATASET_FOO = S3Path("s3:://my_bucket/foo")
const DATASET_FOO_VERSION = "some_identifier"

const DATASET_BAR = S3Path("s3:://my_bucket/bar")
const DATASET_BAR_VERSION = "some_other_identifier"
```

Or a Dictionary, either way that seems bad and like a headache to maintain so I'm all for improving this functionality! I'll take a deeper look at this possibly this afternoon if not tomorrow or early next week.

hannahilea: Hah, exactly---that is exactly what it looks like. :) 

Thank you!

omus:
bors try

src/s3path.jl
```diff
@@ -4,17 +4,18 @@ struct S3Path{A<:AbstractAWSConfig} <: AbstractPath
     drive::String
     isdirectory::Bool
     config::A
+    version::String
```
omus, line 7: Verified strings are appropriate. https://docs.aws.amazon.com/AmazonS3/latest/userguide/versioning-workflows.html#version-ids

Possibly we may want to use `nothing` for unversioned/versioned unaware `S3Path`s

src/s3path.jl
```diff
@@ -123,7 +130,8 @@ function Base.:(==)(a::S3Path, b::S3Path)
 end
 
 function Base.getproperty(fp::S3Path, attr::Symbol)
-    if attr === :anchor
+    if attr 
+      :anchor
```
omus, line 134: Looks to be a mistake:
```suggestion
    if attr === :anchor
```

test/s3path.jl
```diff
@@ -354,4 +355,34 @@ end
     rm(json_path)
 end
 
+@testset "S3Path versioning" begin
+    key_version_file = "test_versions"
+    s3_put(bucket_name, key_version_file, "data.v1")
+    s3_put(bucket_name, key_version_file, "data.v2")
+
+    # `s3_list_versions` returns versions in the order newest to oldest
+    versions = reverse!(s3_list_versions(aws, bucket_name, key_version_file))
```
omus, line 364: Should simplify the remainder of the tests
```suggestion
    versions = reverse(d["VersionId"] for d in s3_list_versions(aws, bucket_name, key_version_file))
```

src/s3path.jl
omus: After reading over the entire thing I think using `nothing` is a better option
```suggestion
    version::Union{String,Nothing}
```
You may want to have:
```julia
const S3PathVersion = Union{String,Nothing}
```

hannahilea: Updated!

omus:
bors try

test/s3path.jl
```diff
@@ -354,4 +355,34 @@ end
     rm(json_path)
 end
 
+@testset "S3Path versioning" begin
+    key_version_file = "test_versions"
+    s3_put(bucket_name, key_version_file, "data.v1")
+    s3_put(bucket_name, key_version_file, "data.v2")
+
+    # `s3_list_versions` returns versions in the order newest to oldest
+    versions = reverse(d["VersionId"] for d in s3_list_versions(aws, bucket_name, key_version_file))
```
omus, line 364: My bad: `reverse` doesn't work on generators
```suggestion
    versions = [d["VersionId"] for d in reverse!(s3_list_versions(aws, bucket_name, key_version_file))]
```

omus:
bors try

src/AWSS3.jl
```diff
@@ -115,7 +116,7 @@ s3_get(a...; b...) = s3_get(global_aws_config(), a...; b...)
 Like `s3_get` but streams result directly to `filename`.  Keyword arguments accept are
 the same as those for `s3_get`.
 """
-function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version="", kwargs...)
+function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version::S3PathVersion=nothing, kwargs...)
     stream = s3_get(aws, bucket, path; return_stream=true, kwargs...)
```
omus, line 120: Currently this function ignores `version` entirely:
```suggestion
function s3_get_file(aws::AbstractAWSConfig, bucket, path, filename; version::S3PathVersion=nothing, kwargs...)
    stream = s3_get(aws, bucket, path; return_stream=true; version=version, kwargs...)
```
Should add a test for this using a versioned file

src/AWSS3.jl
```diff
@@ -85,7 +86,7 @@ function s3_get(
             "return_stream" => return_stream,
         )
 
-        if !isempty(version)
+        if !isnothing(version)
```
omus, line 89: Making this change non-breaking
```suggestion
        if !isnothing(version) || !isempty(version)
```

src/AWSS3.jl
```diff
@@ -147,8 +149,9 @@ end
 
 Retrieves metadata from an object without returning the object itself.
 """
-function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version="", kwargs...)
-    if isempty(version)
+function s3_get_meta(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
+                     kwargs...)
+    if isnothing(version)
```
omus, line 154: Backwards compat
```suggestion
    if isnothing(version) || isempty(version)
```

src/AWSS3.jl
```diff
@@ -187,8 +190,9 @@ s3_exists(a...; b...) = s3_exists(global_aws_config(), a...; b...)
 
 [DELETE Object](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectDELETE.html)
 """
-function s3_delete(aws::AbstractAWSConfig, bucket, path; version="", kwargs...)
-    if isempty(version)
+function s3_delete(aws::AbstractAWSConfig, bucket, path; version::S3PathVersion=nothing,
+                   kwargs...)
+    if isnothing(version)
```
omus, line 195:
```suggestion
    if isnothing(version) || isempty(version)
```

src/s3path.jl
```diff
@@ -4,17 +4,18 @@ struct S3Path{A<:AbstractAWSConfig} <: AbstractPath
...
     isdirectory::Bool
     config::A
+    version::S3PathVersion
 end
 
 # constructor that converts but does not require type parameter
 function S3Path(segments, root::AbstractString, drive::AbstractString, isdirectory::Bool,
-                config::AbstractAWSConfig)
-    S3Path{typeof(config)}(segments, root, drive, isdirectory, config)
+                config::AbstractAWSConfig, version::S3PathVersion)
```
omus, line 12: Backwards compat:
```suggestion
                config::AbstractAWSConfig, version::S3PathVersion=nothing)
```

src/s3path.jl
```diff
@@ -43,6 +45,7 @@ function S3Path()
         "s3://$account_id-$region",
         true,
         config,
+        nothing
```
omus, line 48:
```suggestion
        nothing,
```

src/s3path.jl
```diff
@@ -53,13 +56,15 @@ function S3Path(
     key::AbstractString;
     isdirectory::Bool=false,
     config::AbstractAWSConfig=global_aws_config(),
+    version::S3PathVersion=nothing
```
omus, line 59:
```suggestion
    version::S3PathVersion=nothing,
```

src/s3path.jl
```diff
@@ -53,13 +56,15 @@ function S3Path(
...
 )
     return S3Path(
         Tuple(filter!(!isempty, split(key, "/"))),
         "/",
         strip(startswith(bucket, "s3://") ? bucket : "s3://$bucket", '/'),
         isdirectory,
         config,
+        version
```
omus, line 67:
```suggestion
        version,
```

src/s3path.jl
```diff
@@ -68,25 +73,27 @@ function S3Path(
     key::AbstractPath;
     isdirectory::Bool=false,
     config::AbstractAWSConfig=global_aws_config(),
+    version::S3PathVersion=nothing
```
omus, line 76:
```suggestion
    version::S3PathVersion=nothing,
```

src/s3path.jl
```diff
@@ -68,25 +73,27 @@ function S3Path(
...
 )
     return S3Path(
         key.segments,
         "/",
         normalize_bucket_name(bucket),
         isdirectory,
         config,
+        version
```
omus, line 84:
```suggestion
        version,
```

omus (changes requested):

omus:
bors try

src/AWSS3.jl
hannahilea: yikes! added this (and a test)...

omus:
bors try

src/s3path.jl
```diff
@@ -68,25 +73,27 @@ function S3Path(
...
         config,
+        version,
     )
 end
 
 # To avoid a breaking change.
-function S3Path(str::AbstractString; config::AbstractAWSConfig=global_aws_config())
-    result = tryparse(S3Path, str; config=config)
+function S3Path(str::AbstractString; config::AbstractAWSConfig=global_aws_config(), version=nothing)
+    result = tryparse(S3Path, str; config=config, version=version)
```
omus, line 90: Based on the comment above the function we should probably just not update this as it was unofficially deprecated in https://github.com/JuliaCloud/AWSS3.jl/pull/88.
```suggestion
function S3Path(str::AbstractString; config::AbstractAWSConfig=global_aws_config())
    result = tryparse(S3Path, str; config=config)
```

src/s3path.jl
```diff
@@ -68,25 +73,27 @@ function S3Path(
...
     result !== nothing || throw(ArgumentError("Invalid s3 path string: $str"))
     return result
 end
 
 # if config=nothing, will not try to talk to AWS until after string is confirmed to be an s3 path
-function Base.tryparse(::Type{<:S3Path}, str::AbstractString; config::Union{Nothing,AbstractAWSConfig}=nothing)
+function Base.tryparse(::Type{<:S3Path}, str::AbstractString; config::Union{Nothing,AbstractAWSConfig}=nothing, version::S3PathVersion=nothing)
```
omus, line 96: I feel like in practise having the version as part of the parsed string would be the most useful. We could probably use the HTTP GET parameter format which is already used by S3 HTTP requests: https://docs.aws.amazon.com/AmazonS3/latest/userguide/RetMetaOfObjVersion.html

Probably the easiest approach to this is to leave out `version` support here unless you need this right away and we can figure out the right approach independent from the rest of the support.

src/s3path.jl
```diff
@@ -158,6 +165,7 @@ function FilePathsBase.join(prefix::S3Path, pieces::AbstractString...)
         prefix.drive,
         isdirectory,
         prefix.config,
+        nothing, # Version is per-object, so would not propogate from the prefix
```
omus, line 168:
```suggestion
        nothing, # Version is per-object, so we should not propagate it from the prefix
```

src/s3path.jl
```diff
@@ -173,7 +181,7 @@ function FilePathsBase.parents(fp::S3Path)
     end
 end
 
-FilePathsBase.exists(fp::S3Path) = s3_exists(fp.config, fp.bucket, fp.key)
+FilePathsBase.exists(fp::S3Path) = s3_exists(fp.config, fp.bucket, fp.key; version=fp.version)
```
omus, line 184: Interesting tidbit I just discovered: this can be `s3_exists(fp.config, fp.bucket, fp.key; fp.version)`
```julia
julia> struct Demo
           x
           y
       end

julia> d = Demo(1,2)
Demo(1, 2)

julia> f(; y) = y
f (generic function with 1 method)

julia> f(; d.y)
2
```

src/AWSS3.jl
```diff
@@ -36,6 +36,7 @@ using UUIDs
 @service S3
 
 const SSDict = Dict{String,String}
+const S3PathVersion = Union{String,Nothing}
```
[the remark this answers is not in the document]
omus: We could do:
```julia
const AbstractS3PathVersion = Union{AbstractString,Nothing}
const S3PathVersion = Union{String,Nothing}
```
but I think that's probably overkill

src/s3path.jl
omus: I've definitely changed my mind and do think `version` makes sense to be part of the `S3Path`. There is some weirdness here but this PR is a step in the right direction.

omus:
Thanks @hannahilea for sticking with this. The change from `""` to `nothing` was more work than I expected but it did uncover a pre-existing bug. Should be the last set of changes needed to get this merged.

src/s3path.jl
ericphanson: does mean the non-"deprecated" way is to use the `S3Path(bucket, key)` constructor? Because I've actually never seen that constructor used before, only this `S3Path("s3://bucket/key")` version. So I'd be against the soft-deprecation-- feels disruptive for no reason.

src/s3path.jl
ericphanson: only in Julia 1.5+ unforunately, I love this syntax 😄

src/s3path.jl
ericphanson: ah, I see this is probably so that we don't need to pass it to `tryparse` to instead parse it out of a string? I guess that makes sense

omus: I believe `parse(S3Path, str)` is probably the preferred way to do this. @rofinn can comment further

src/s3path.jl
omus: FYI: https://github.com/JuliaLang/Compat.jl/pull/747. I need this for something else

src/s3path.jl
hannahilea: Oh interesting. I agree that we should additionally support having the version as part of the parsed string, but I think having this current implementation is orthogonally important. 

E.g., without it, we're back to needing to construct all of our paths like so:
```
my_path = S3Path("my_bucket", "my_key"; version="my_version")
```
instead of
```
my_path = S3Path("s3://my_bucket/my_key"; version="my_version")
```
which makes it a (slight) pain to pass around variables-per-object.  

What would you think about keeping the current implementation here and then (in a separate PR, I can open an issue) additionally supporting parsing the version _from_ `str`. (Probably something like 
```
str = "s3://my_bucket/my_key?versionId=L4kqtJlcpXroDTDmpUMLUo"
```
, a la the REST API command: https://docs.aws.amazon.com/AmazonS3/latest/userguide/RetrievingObjectVersions.html)

...I also think this would be straightforward enough to add/add tests for here.

omus: I'll call out I don't think
```julia
S3Path("s3://my_bucket/my_key"; version="my_version")
```
currently works since the above `S3Path` constructor doesn't take a `version` keyword: https://github.com/JuliaCloud/AWSS3.jl/pull/161/files#diff-31b593db3c82f33730f4af58846ed59213199a8dd973bddc4f70f1e1721d6719R89

Maybe a good path forward is to:
- Add the `version` keyword to `S3Path(::AbstractString; config::AbstractAWSConfig)`
- Remove the `version` keyword from `tryparse`. This avoids having to deprecate the keyword here in the future

To do this we'd need to either make an `S3Path` without a version and then make a new one with the version added in or revise `tryparse` to call a private function that returns the parsed components

omus:
bors try

omus:
bors try

src/s3path.jl
hannahilea: > I'll call out I don't think 
> [...]
> currently works since the above S3Path constructor doesn't take a version keyword:

...ah right, b/c we literally just removed it. :| 

> Maybe a good path forward is to:
> - Add the version keyword to S3Path(::AbstractString; config::AbstractAWSConfig)
> - Remove the version keyword from tryparse. This avoids having to deprecate the keyword in the future for tryparse
To do this we'd need to either make an S3Path without a version and then make a new one with the version added in or revise tryparse to call a private function that returns the parsed components

Yeah, this sounds good to me, I'll make that change. I think prefer your first approach, so I'll go with that.

hannahilea: Righto, change made---see what you think. For future safety, I threw in an error in the case that both a version and a parseable-string-containing-a-version are passed in, although we won't ever hit that error until we do the follow-up work of parsing the version from the string in the first place.

ericphanson:
bors try

ericphanson:
@hannahilea this should let you run CI yourself so you don't have to wait for others to run it:

bors delegate+

hannahilea:
bors try

hannahilea:
bors try

omus:
bors try

test/s3path.jl
```diff
@@ -354,4 +355,39 @@ end
...
+    unversioned_path = S3Path(bucket_name, key_version_file; config=aws)
+    versioned_path = S3Path(bucket_name, key_version_file; config=aws, version=last(versions))
+    @test versioned_path.version == last(versions)
+    @test unversioned_path.version === nothing
+    @test exists(versioned_path)
+    @test exists(unversioned_path)
+    nonexistent_versioned_path = S3Path(bucket_name, key_version_file; config=aws, version="feVMBvDgNiKSpMS17fKNJK3GV05bl8ir")
+    @test !exists(nonexistent_versioned_path)
+
+    @test S3Path("s3://$(bucket_name)/$(key_version_file)"; version=first(versions)).version == version
```
omus, line 383:
```suggestion
    version = first(versions)
    @test S3Path("s3://$(bucket_name)/$(key_version_file)"; version=version).version == version
```

hannahilea:
bors try

test/s3path.jl
hannahilea: ...jinx, already fixed test. :)

test/s3path.jl
```diff
@@ -354,4 +355,39 @@ end
...
+
+    # `s3_list_versions` returns versions in the order newest to oldest
+    versions = [d["VersionId"] for d in reverse!(s3_list_versions(aws, bucket_name, key_version_file))]
+    @test length(versions) == 2
+    @test read(S3Path(bucket_name, key_version_file; config=aws, version=first(versions)), String) == "data.v1"
+    @test read(S3Path(bucket_name, key_version_file; config=aws, version=last(versions)), String) == "data.v2"
+    @test isequal(read(S3Path(bucket_name, key_version_file; config=aws, version=last(versions)), String),
+                  read(S3Path(bucket_name, key_version_file; config=aws), String))
+    @test isequal(read(S3Path(bucket_name, key_version_file; config=aws, version=last(versions)), String),
+                  read(S3Path(bucket_name, key_version_file; config=aws, version=""), String))
```
omus, line 372: Using `nothing` should be the preferred way to get the latest version
```suggestion
                  read(S3Path(bucket_name, key_version_file; config=aws, version=nothing), String))
```

src/AWSS3.jl
```diff
@@ -36,6 +36,7 @@ using UUIDs
 @service S3
 
 const SSDict = Dict{String,String}
+const S3PathVersion = Union{AbstractString,Nothing}
```
omus, line 39:
```suggestion
const AbstractS3Version = Union{AbstractString,Nothing}
```

omus:
If this PR is included in the breaking release that includes #175 we may also want to do: https://github.com/JuliaCloud/AWSS3.jl/issues/181

omus:
bors try

src/s3path.jl
```diff
@@ -95,7 +107,8 @@ function Base.tryparse(::Type{<:S3Path}, str::AbstractString; config::Union{Noth
     path = ()
     isdirectory = true
 
-    tokenized = split(str, "/")
+    version_split = split(str, "?version=")
```
omus, line 110: Let's stick to using `versionId` to match https://docs.aws.amazon.com/AmazonS3/latest/userguide/RetrievingObjectVersions.html#retrieve-obj-version-rest
```suggestion
    version_split = split(str, "?versionId=")
```
The parsing is frail and should probably use URIs.jl but I think it's adequate for the immediate future

hannahilea: ...I'm going to revert this chunk, I didn't actually intend to push this change on this branch! Will open in a separate PR, as originally intended (and will update to `versionId` there).

hannahilea:
bors try

hannahilea:
bors try

test/awss3.jl
```diff
@@ -126,6 +126,10 @@ end
     @test (s3_get(aws, bucket_name, "key3"; version=versions[3]["VersionId"]) == b"data3.v1")
     @test (s3_get(aws, bucket_name, "key3"; version=versions[2]["VersionId"]) == b"data3.v2")
     @test (s3_get(aws, bucket_name, "key3"; version=versions[1]["VersionId"]) == b"data3.v3")
+
+    tmp_file = "/tmp/jl_qws_test_key3"
+    s3_get_file(aws, bucket_name, "key3", tmp_file; version=versions[2]["VersionId"])
+    @test (read(tmp_file) == b"data3.v2")
```
omus, line 132:
```suggestion
    @test read(tmp_file) == b"data3.v2"
```

src/s3path.jl
omus: Sounds great. Definitely not a critical feature for this PR and makes sense to be separate

test/s3path.jl
```diff
@@ -354,4 +355,39 @@ end
...
+    unversioned_path = S3Path(bucket_name, key_version_file; config=aws)
+    versioned_path = S3Path(bucket_name, key_version_file; config=aws, version=last(versions))
+    @test versioned_path.version == last(versions)
+    @test unversioned_path.version === nothing
+    @test exists(versioned_path)
+    @test exists(unversioned_path)
+    nonexistent_versioned_path = S3Path(bucket_name, key_version_file; config=aws, version="feVMBvDgNiKSpMS17fKNJK3GV05bl8ir")
+    @test !exists(nonexistent_versioned_path)
+
+    @test S3Path("s3://$(bucket_name)/$(key_version_file)"; version=first(versions)).version == first(versions)
```
omus, line 383: Something about this makes my eyes bleed
```suggestion
    versioned_path_v1 = S3Path("s3://$(bucket_name)/$(key_version_file)"; version=first(versions)
    @test versioned_path_v1.version == first(versions)
```

omus (approved):
image

Last suggestions are superficial

hannahilea:
bors try

hannahilea:
bors try

ericphanson:
If I understand things right, the current state allows both empty strings and nothing, so is purely non-breaking/non-deprecating and we can merge this now and tag a v0.8-series release with it? I've tagged #181 for v0.9 so we can keep that separate from this PR.

ericphanson:
looks like db41c2b35ff5ac9e3a7dd671908e89e78654819a wasn't quite right so I reverted it for now.

bors try

test/awss3.jl
```diff
@@ -126,6 +126,10 @@ end
     @test (s3_get(aws, bucket_name, "key3"; version=versions[3]["VersionId"]) == b"data3.v1")
     @test (s3_get(aws, bucket_name, "key3"; version=versions[2]["VersionId"]) == b"data3.v2")
     @test (s3_get(aws, bucket_name, "key3"; version=versions[1]["VersionId"]) == b"data3.v3")
+
+    tmp_file = "/tmp/jl_qws_test_key3"
```
omus, line 130: Avoid making assumptions about the filesystem:
```suggestion
    tmp_file = joinpath(tempdir(), "jl_qws_test_key3")
```

omus:
@hannahilea mentioned to me she was going to add version equivalence checks before merging

hannahilea:
bors try

PR #162, opened 2021-06-18
Fix `read(::S3Path)` for e.g. JSON

### LINKED ISSUE #138: `read(path::S3Path)` fails when reading JSON
to reproduce, simply put a `.json` file in an S3 bucket, and make an S3Path pointing to it, and try to `read` it. The problem is https://github.com/JuliaCloud/AWS.jl/issues/224, i.e. AWS.jl decides to return a dictionary instead of the raw bytes, and the `Vector{UInt8}` constructor fails in 
https://github.com/JuliaCloud/AWSS3.jl/blob/3424fe400b98d24591cb4d1915846e6204b542bc/src/s3path.jl#L347

Maybe this is a good time to mention I really strongly disagree with heuristically deciding what type to return from `get_object` instead of just always returning the raw bytes.

ericphanson:
fixes #138 

This was defined as
```julia
Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key))
```
which fails whenever `s3_get` does not return something that can pass to `Vector{UInt8}`, e.g. a LittleDict. Here we add `raw=true` to fix this.

Co-authored by @hannahilea

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "AWSS3"
 uuid = "1c724243-ef5b-51ab-93f4-b0a88ac62a95"
-version = "0.8.3"
+version = "0.8.4"
 
 [deps]
 AWS = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
@@ -21,6 +21,7 @@ AWS = "1.25"
 EzXML = "0.9, 1"
 FilePathsBase = "0.9"
 HTTP = "0.8, 0.9"
+JSON3 = "1"
 MbedTLS = "0.6, 0.7, 1"
 OrderedCollections = "1"
 Retry = "0.3, 0.4"
@@ -29,7 +30,8 @@ XMLDict = "0.3, 0.4"
 julia = "1"
 
 [extras]
+JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test"]
+test = ["Test", "JSON3"]
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -527,7 +527,7 @@ end
 s3_purge_versions(a...; b...) = s3_purge_versions(global_aws_config(), a...; b...)
 
 """
-    s3_put([::AbstractAWSConfig], bucket, path, data; <keyword arguments>)
+    s3_put([::AbstractAWSConfig], bucket, path, data, data_type="", encoding=""; <keyword arguments>)
 
 [PUT Object](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html)
 `data` at `path` in `bucket`.
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -355,7 +355,7 @@ function Base.readdir(fp::S3Path; join=false, sort=true)
     end
 end
 
-Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key))
+Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key; raw=true))
 
 Base.write(fp::S3Path, content::String; kwargs...) = Base.write(fp, Vector{UInt8}(content); kwargs...)
 
--- a/test/s3path.jl
+++ b/test/s3path.jl
@@ -337,6 +337,9 @@ end
 @testset "JSON roundtripping" begin
     json_path = S3Path("s3://$(bucket_name)/test_json"; config=aws)
     my_dict = Dict("key" => "value", "key2" => 5.0)
+    # here we use the "application/json" MIME type to trigger the heuristic parsing into a `LittleDict`
+    # that will hit a `MethodError` at the `Vector{UInt8}` constructor of `read(::S3Path)` if `raw=true`
+    # was not passed to `s3_get` in that method.
     s3_put(aws, bucket_name, "test_json", JSON3.write(my_dict), "application/json")
     json_bytes = read(json_path)
     @test JSON3.read(json_bytes, Dict) == my_dict
--- a/test/s3path.jl
+++ b/test/s3path.jl
@@ -334,4 +334,13 @@ end
     end
 end
 
+@testset "JSON roundtripping" begin
+    json_path = S3Path("s3://$(bucket_name)/test_json"; config=aws)
+    my_dict = Dict("key" => "value", "key2" => 5.0)
+    s3_put(aws, bucket_name, "test_json", JSON3.write(my_dict), "application/json")
+    json_bytes = read(json_path)
+    @test JSON3.read(json_bytes, Dict) == my_dict
+    rm(json_path)
+end
+
 AWSS3.s3_nuke_bucket(aws, bucket_name)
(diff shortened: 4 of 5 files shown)


mattBrzezinski:
bors try

ericphanson:
This changes what happens if you have the wrong region. Before, you'd get a confusing LittleDict error. Now, you get some bytes back that correspond to the string
```julia
"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>PermanentRedirect</Code><Message>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.</Message>...[truncated]"
```
which I truncated to remove the bucket and key info. I think an exception should be raised, but I'm not sure the right level and where this should be realized. I've asked @omus to help me figure that out 🙂

mattBrzezinski:
> This changes what happens if you have the wrong region. Before, you'd get a confusing LittleDict error. Now, you get some bytes back that correspond to the string
> 
> ```julia
> "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>PermanentRedirect</Code><Message>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.</Message>...[truncated]"
> ```
> 
> which I truncated to remove the bucket and key info. I think an exception should be raised, but I'm not sure the right level and where this should be realized. I've asked @omus to help me figure that out 🙂

You would probably want to add in a clause [here](https://github.com/JuliaCloud/AWSS3.jl/blob/2b564443281d72c5520cbc3a156e479c35d9ee2a/src/AWSS3.jl#L86).

ericphanson:
> You would probably want to add in a clause [here](https://github.com/JuliaCloud/AWSS3.jl/blob/2b564443281d72c5520cbc3a156e479c35d9ee2a/src/AWSS3.jl#L86).

I don't think that will change anything; `S3.get_object` is not throwing anything in this case:
```julia
julia> S3.get_object(bucket, path, Dict("return_raw" => true); aws_config=aws)
483-element Vector{UInt8}:
...
```
which is the same response as above. In fact, nothing throws; the HTTP response we get back is:
```julia
HTTP.Messages.Response:
"""
HTTP/1.1 301 Moved Permanently
x-amz-bucket-region: us-east-2
x-amz-request-id: XXX
x-amz-id-2: XXX
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Mon, 21 Jun 2021 10:34:53 GMT
Server: AmazonS3

<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>PermanentRedirect</Code><Message>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.</Message><Endpoint>XXX.s3.us-east-2.amazonaws.com</Endpoint><Bucket>XXX</Bucket><RequestId>XXX</RequestId><HostId>XXX</HostId></Error>"""
```
It seems `HTTP.request` does not throw, even though `status_exception=true`, and `response.status > 300`, because `HTTP.Messages.isredirect(response)` is true (which makes sense, it is a redirect saying we're in the wrong bucket).

So in other words, we're getting a response back that it is a redirect, but not handling that (though I don't know that we'd want to automatically handle that) or throwing, so we end up sending the message all the way back to the caller.

I think maybe `submit_request` is the right place to handle this, though I'm not sure.

ericphanson:
This is good to go from my point of view, since the concern I brought up in https://github.com/JuliaCloud/AWSS3.jl/pull/162#issuecomment-864150746 is an orthogonal bug which is fixed by https://github.com/JuliaCloud/AWS.jl/pull/389

ericphanson:
Bump

src/AWSS3.jl
```diff
@@ -527,7 +527,7 @@ end
 s3_purge_versions(a...; b...) = s3_purge_versions(global_aws_config(), a...; b...)
 
 """
-    s3_put([::AbstractAWSConfig], bucket, path, data; <keyword arguments>)
+    s3_put([::AbstractAWSConfig], bucket, path, data, data_type="", encoding=""; <keyword arguments>)
```
ericphanson, line 530: these were missing from the docstring, which made it seem like they were keyword arguments

src/s3path.jl
```diff
@@ -355,7 +355,7 @@ function Base.readdir(fp::S3Path; join=false, sort=true)
     end
 end
 
-Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key))
+Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key; raw=true))
```
ericphanson, line 358: this is the only functional change here, and due to the `Vector{UInt8}` constructor, if one was not getting back raw bytes here, they would get a confusing `MethodError` from the constructor. Now they will get the bytes back.

We saw this often previously when AWS.jl would not throw on redirects and instead send the content of the redirect message back, which would get converted to a LittleDict and error at this point. However, this constructor is not the correct place to check for these things; errors associated to non-OK HTTP status codes should be thrown in `submit_request` with the current design of AWS.jl.

test/s3path.jl
```diff
@@ -334,4 +334,13 @@ end
     end
 end
 
+@testset "JSON roundtripping" begin
+    json_path = S3Path("s3://$(bucket_name)/test_json"; config=aws)
+    my_dict = Dict("key" => "value", "key2" => 5.0)
+    s3_put(aws, bucket_name, "test_json", JSON3.write(my_dict), "application/json")
```
ericphanson, line 340: here we use the `"application/json"` MIME type to trigger the heuristic parsing into a LittleDict that will hit a `MethodError` at the `Vector{UInt8}` constructor.

test/s3path.jl
```diff
@@ -334,4 +334,13 @@ end
...
+    json_bytes = read(json_path)
+    @test JSON3.read(json_bytes, Dict) == my_dict
+    rm(json_path)
```
ericphanson, line 343: since we are in a `@testset`, this line should get called even if a previous line throws. Besides, we nuke the bucket at the end anyway.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "AWSS3"
 uuid = "1c724243-ef5b-51ab-93f4-b0a88ac62a95"
-version = "0.8.3"
+version = "0.8.4"
```
ericphanson, line 3: this is a bugfix so we bump the patch version

test/s3path.jl
omus: If an exception is raised then the `@testset` will be aborted early.

test/s3path.jl
omus: I think it is worth adding this as comment here

omus (approved):

test/s3path.jl
ericphanson: oh really? good to know!

test/s3path.jl
ericphanson, line 340:
```suggestion
    # here we use the "application/json" MIME type to trigger the heuristic parsing into a `LittleDict`
    # that will hit a `MethodError` at the `Vector{UInt8}` constructor of `read(::S3Path)` if `raw=true`
    # was not passed to `s3_get` in that method.
    s3_put(aws, bucket_name, "test_json", JSON3.write(my_dict), "application/json")
```

ericphanson: added!

mattBrzezinski (approved):
bors r+

Issue #147, opened 2021-03-01
Slow performance of `readdir(::S3Path)`


omus:
Using `aws s3 ls s3://private-bucket/example/gz/2018-06-29/` takes 2 seconds to return a list of 9 keys. The equivalent Julia command takes far longer:

```julia
julia> @time readdir(S3Path("s3://private-bucket/example/gz/2018-06-29/"))
371.723262 seconds (93.69 M allocations: 4.317 GiB, 0.33% gc time)
```
From looking at the source code it appears that we're [retrieving all S3 paths with the specified prefix](https://github.com/JuliaCloud/AWSS3.jl/blob/master/src/s3path.jl#L343) effectively doing a recursive search and then filtering the results.

omus:
cc: @cbdavis

Issue #158, opened 2021-05-13
Walkpath is slow, we can make it easily 10x faster.


oxinabox:
Related to #147 

Copying this over from some internal code @rofinn  wrote, so that this doesn't get lost and forgotten entirely.

> Slight optimization over the default `readdir(::S3Path)` based method to avoid multiple
calls when we want to grab all prefixes at once and we don't care about order:
> we may need tests or custom logic to ensure that ordering is consistent.

```julia
function _walkpath(fp::S3Path)
    # Select objects within that prefix
    objects = s3_list_objects(fp.config, fp.bucket, fp.key; delimiter="")

    # Extract keys from the results and strip the prefix out
    keys = [lstrip(replace(o["Key"], fp.key => ""), '/') for o in objects]

    # Filter any empty keys (e.g, prefix)
    filter!(!isempty, keys)

    # Return the keys joined with the original prefix
    return joinpath.(fp, keys)
end
```

Benchmark

```julia
julia> using FilePathsBase, AWSS3, BenchmarkTools

julia> fp = S3Path("REDACTED");

julia> @btime collect(walkpath($fp));
  465.464 ms (10916 allocations: 635.63 KiB)

julia> @btime collect(_walkpath($fp));
  66.557 ms (2430 allocations: 131.33 KiB)
```

ericphanson:
Just to say

> `replace(o["Key"], fp.key => "")`

is the correctness bug I fixed in #164; probably want `chop` like in https://github.com/JuliaCloud/AWSS3.jl/pull/164/files#diff-31b593db3c82f33730f4af58846ed59213199a8dd973bddc4f70f1e1721d6719R346.

PR #164, opened 2021-06-22
greatly improve `readdir` perf; fix bug

ericphanson:
fixes https://github.com/JuliaCloud/AWSS3.jl/issues/147

Currently, we iterate over all objects with a given prefix, even deeply nested ones; in other words, we essentially do a full `walkpath`, and then truncate the keys to just one level and `unique` them. This is very slow. This makes the default `walkpath` implementation super duper slow, because it repeats an entire `walkpath` for every level of the directory hierarchy! ref #158. This PR will definitely improve performance there, but probably just adding a direct `walkpath` implementation would be even faster.

However, if you pass a `delimiter`, then AWS will just return prefixes up to that delimiter (see e.g. the example "Sample Request: Using the delimiter and prefix parameters"  from https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html#API_ListObjectVersions_RequestSyntax). This is the quick way to do it. It returns both a "Contents" field if there are any objects at that level, plus a "CommonPrefix" field, with the "directories". Our `s3_list_objects` method ignores the "CommonPrefix", which is probably why the current implementation does not pass a delimeter. `s3_list_objects` is also documented to return various pieces of metadata which we don't have for directories listed in the "CommonPrefix".

Instead, I just directly call the s3 API in the new readdir method. I use the `list_objects_v2` API since the continuation tokens are easier to handle than the `marker` system.

This also fixes a small subtle bug: the `s3_get_name` method used in the previous version calls `replace` to remove the key from the start of the returned prefix. This means if we had an object like `s3://bucket/A/A/B.txt` and call `readdir(S3Path("s3://bucket/A/"))` we would get back `B.txt` instead of `A/`.

I added a test for this, and checked locally that it failed on the old version of `readdir` (using Minio):
```julia
julia> path = S3Path("s3://$(bucket_name)/A/A/B.txt"; config = aws)
p"s3://ocaws.jl.test.20210622t110746z/A/A/B.txt"

julia> write(path, "test!")
UInt8[]

julia> results = readdir(S3Path("s3://$(bucket_name)/A/"; config = aws))
1-element Vector{SubString{String}}:
 "B.txt"

julia> @test results == ["A/"]
Test Failed at REPL[26]:1
  Expression: results == ["A/"]
   Evaluated: SubString{String}["B.txt"] == ["A/"]
ERROR: There was an error during testing
```

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -342,25 +342,23 @@ function _pair_or_dict_get(p::Pair, k)
 end
 _pair_or_dict_get(d::AbstractDict, k) = get(d, k, nothing)
 
-function _readdir_add_results!(results, r, key_length)
+function _retrieve_value!(results, result_key, prefix_key, key_length)
     rm_key = s -> chop(s, head=key_length, tail=0)
+    if haskey(r, result_key)
+        for p in r[result_key]
+            prefix = _pair_or_dict_get(p, prefix_key)
+            if prefix !== nothing
+                push!(results, rm_key(prefix))
+            end
+        end
+    end
+    return nothing
+end
+
+function _readdir_add_results!(results, r, key_length)
     sizehint!(results, length(results) + parse(Int, r["KeyCount"]))
-    if haskey(r, "CommonPrefixes")
-        for p in r["CommonPrefixes"]
-            prefix = _pair_or_dict_get(p, "Prefix")
-            if prefix !== nothing
-                push!(results, rm_key(prefix))
-            end
-        end
-    end
-    if haskey(r, "Contents")
-        for o in r["Contents"]
-            prefix = _pair_or_dict_get(o, "Key")
-            if prefix !== nothing
-                push!(results, rm_key(prefix))
-            end
-        end
-    end
+    _retrieve_value!(results, "CommonPrefixes", "Prefix", key_length)
+    _retrieve_value!(results, "Contents", "Key", key_length)
     return get(r, "NextContinuationToken", nothing)
 end
 
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -359,7 +359,7 @@ function _retrieve_prefixes!(results, objects, prefix_key, chop_head)
 end
 
 function _readdir_add_results!(results, response, key_length)
-    sizehint!(results, length(results) + parse(Int, r["KeyCount"]))
+    sizehint!(results, length(results) + parse(Int, response["KeyCount"]))
 
     _retrieve_prefixes!(results, get(response, "CommonPrefixes", nothing), "Prefix", key_length)
     _retrieve_prefixes!(results, get(response, "Contents", nothing), "Key", key_length)
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -344,14 +344,17 @@ _pair_or_dict_get(d::AbstractDict, k) = get(d, k, nothing)
 
 function _retrieve_value!(results, result_key, prefix_key, key_length)
     rm_key = s -> chop(s, head=key_length, tail=0)
+    
     if haskey(r, result_key)
         for p in r[result_key]
             prefix = _pair_or_dict_get(p, prefix_key)
+            
             if prefix !== nothing
                 push!(results, rm_key(prefix))
             end
         end
     end
+    
     return nothing
 end
 
@@ -359,6 +362,7 @@ function _readdir_add_results!(results, r, key_length)
     sizehint!(results, length(results) + parse(Int, r["KeyCount"]))
     _retrieve_value!(results, "CommonPrefixes", "Prefix", key_length)
     _retrieve_value!(results, "Contents", "Key", key_length)
+
     return get(r, "NextContinuationToken", nothing)
 end
 
@@ -366,13 +370,12 @@ function Base.readdir(fp::S3Path; join=false, sort=true)
     if isdir(fp)
         k = fp.key
         key_length = length(k)
-
         results = String[]
- 
         token = ""
         while token !== nothing
             r = @repeat 4 try
                 params = Dict("delimiter" => "/", "prefix" => k)
+
                 if !isempty(token)
                     params["continuation-token"] = token
                 end
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -364,7 +364,7 @@ function _readdir_add_results!(results, response, key_length)
     _retrieve_prefixes!(results, get(response, "CommonPrefixes", nothing), "Prefix", key_length)
     _retrieve_prefixes!(results, get(response, "Contents", nothing), "Key", key_length)
 
-    return get(r, "NextContinuationToken", nothing)
+    return get(response, "NextContinuationToken", nothing)
 end
 
 function Base.readdir(fp::S3Path; join=false, sort=true)
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -333,14 +333,43 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
     end
 end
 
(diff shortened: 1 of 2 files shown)


src/s3path.jl
```diff
@@ -333,14 +333,57 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
 
-        # Only list the basename and not the full key
-        results = unique!([s3_get_name(k, string(o["Key"])) for o in objects])
+        results = String[]
+        r = @repeat 4 try
+            S3.list_objects_v2(fp.bucket, Dict("delimiter" => "/", "prefix" => k); aws_config=fp.config)
+        catch e
+            @delay_retry if ecode(e) in ["NoSuchBucket"] end
+        end
+        token = _readdir_add_results!(results, r, key_length)
```
omus, line 378: Let's just start `token` off in a way where we always enter the loop once:
```suggestion
        token = ""
```

test/s3path.jl
```diff
@@ -332,6 +332,13 @@ end
         # test trailing slash on final piece is included
         @test p"s3://foo/bar" / "baz/" == p"s3://foo/bar/baz/"
     end
+
+    @testset "`readdir`" begin
```
omus, line 336:
```suggestion
    @testset "readdir" begin
```

omus:
bors try

src/s3path.jl
ericphanson: hm, is passing `continuation-token=""` the same as not passing it at all? that was my concern about something like that

ericphanson: I get `ERROR: AWSException: InvalidArgument -- The continuation token provided is incorrect` when I try that locally. but I could unify them a bit more

mattBrzezinski:
bors try

ericphanson:
Bump

src/s3path.jl
```diff
@@ -333,14 +333,57 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
+function _pair_or_dict_get(p::Pair, k)
+    first(p) == k || return nothing
+    return last(p)
+end
+_pair_or_dict_get(d::AbstractDict, k) = get(d, k, nothing)
+
+function _readdir_add_results!(results, r, key_length)
+    rm_key = s -> chop(s, head=key_length, tail=0)
+    sizehint!(results, length(results) + parse(Int, r["KeyCount"]))
+    if haskey(r, "CommonPrefixes")
```
mattBrzezinski, line 348: These two `if` statements are just duplicates, we can pull them into their own function. Maybe something like:

```julia
_retrieve_value(results, result_key, prefix_key)
```

ericphanson: sure, done

src/s3path.jl
```diff
@@ -333,14 +333,55 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
+    if haskey(r, result_key)
+        for p in r[result_key]
+            prefix = _pair_or_dict_get(p, prefix_key)
+            if prefix !== nothing
+                push!(results, rm_key(prefix))
+            end
+        end
+    end
+    return nothing
+end
```
mattBrzezinski, line 356:
```suggestion
function _retrieve_value!(results, result_key, prefix_key, key_length)
    rm_key = s -> chop(s, head=key_length, tail=0)
    
    if haskey(r, result_key)
        for p in r[result_key]
            prefix = _pair_or_dict_get(p, prefix_key)
            
            if prefix !== nothing
                push!(results, rm_key(prefix))
            end
        end
    end
    
    return nothing
end
```

src/s3path.jl
```diff
@@ -333,14 +333,55 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
+
+function _readdir_add_results!(results, r, key_length)
+    sizehint!(results, length(results) + parse(Int, r["KeyCount"]))
+    _retrieve_value!(results, "CommonPrefixes", "Prefix", key_length)
+    _retrieve_value!(results, "Contents", "Key", key_length)
+    return get(r, "NextContinuationToken", nothing)
```
mattBrzezinski, line 362:
```suggestion

    return get(r, "NextContinuationToken", nothing)
```

src/s3path.jl
```diff
@@ -333,14 +333,55 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
         k = fp.key
-        # Only list the files and "dirs" within this S3 "dir"
-        objects = s3_list_objects(fp.config, fp.bucket, k; delimiter="")
-
-        # Only list the basename and not the full key
-        results = unique!([s3_get_name(k, string(o["Key"])) for o in objects])
+        key_length = length(k)
+
+        results = String[]
+
```
mattBrzezinski, line 371:
```suggestion
```

src/s3path.jl
```diff
@@ -333,14 +333,55 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
-        results = unique!([s3_get_name(k, string(o["Key"])) for o in objects])
+        key_length = length(k)
+
+        results = String[]
+ 
+        token = ""
+        while token !== nothing
+            r = @repeat 4 try
+                params = Dict("delimiter" => "/", "prefix" => k)
+                if !isempty(token)
```
mattBrzezinski, line 376:
```suggestion

                if !isempty(token)
```

test/s3path.jl
```diff
@@ -332,6 +332,13 @@ end
         # test trailing slash on final piece is included
         @test p"s3://foo/bar" / "baz/" == p"s3://foo/bar/baz/"
     end
+
+    @testset "readdir" begin
+        path = S3Path("s3://$(bucket_name)/A/A/B.txt"; config = aws)
+        write(path, "test!")
+        results = readdir(S3Path("s3://$(bucket_name)/A/"; config = aws))
+        @test results == ["A/"]
```
mattBrzezinski, line 340:
```suggestion

        @test results == ["A/"]
```

mattBrzezinski (approved):
Changes LGTM! Just some minor whitespacing / formatting things. I should really just add in a JuliaFormatter action into this repo hehe...

To explain my possible insanity around the white spacing, it's really just to increase readability of the code.

I find that having a newline before any `return` or `@test` cases makes it really quick to scan around the file, see what the test case is doing.

In function blocks, having like-wise operations (variable assignment) in a grouping, and then separating "blocks" of code (`if statement`, `do-while`, etc) with newlines before and after make it a lot easier on the eyes.

ericphanson:
Looks like I forgot to pass `r` along to the `_retrieve_values!` function. I decided the variable names weren't very clear so I tried to clean it up a bit while fixing that.

src/s3path.jl
```diff
@@ -333,14 +333,59 @@ function FilePathsBase.sync(f::Function, src::AbstractPath, dst::S3Path; delete=
...
+end
+
+function _readdir_add_results!(results, response, key_length)
+    sizehint!(results, length(results) + parse(Int, response["KeyCount"]))
+
+    _retrieve_prefixes!(results, get(response, "CommonPrefixes", nothing), "Prefix", key_length)
+    _retrieve_prefixes!(results, get(response, "Contents", nothing), "Key", key_length)
+
+    return get(r, "NextContinuationToken", nothing)
+end
```
mattBrzezinski, line 368: Can't leave a comment on the line above :/

We're trying to `get(r)` but it DNE?

ericphanson: ohh whoops, that should be `response` as well, sorry. I'll push a fix

PR #167, opened 2021-06-30
make it easier to access by byte range

ExpandingMan:
I somehow did not even know this was a thing even though it's a crucial feature.  Amazingly, there was already a unit test for it.

This PR makes it much nicer for users to specify a byte range and should be non-breaking.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -64,18 +64,23 @@ from `path` in `bucket`.
                 (by default return type depends on `Content-Type` header).
 - `byte_range=nothing`:  given an iterator of `(start_byte, end_byte)` gets only
     the range of bytes of the object from `start_byte` to `end_byte`.  For example,
-    `byte_range=(0,3)` gets bytes 0 to 3 icnlusive.
+    `byte_range=1:4` gets bytes 1 to 4 inclusive.  Arguments should use the Julia convention
+    of 1-based indexing.
 - `header::Dict{String,String}`: pass in an HTTP header to the request.
 
-For example, to get a `range` of bytes instead of the whole object, do:
+As an example of how to set custom HTTP headers, the below is equivalent to
+`s3_get(aws, bucket, path; byte_range=range)`.
 
 `s3_get(aws, bucket, path; headers=Dict{String,String}("Range" => "bytes=\$(first(range)-1)-\$(last(range)-1)"))`
 """
 function s3_get(aws::AbstractAWSConfig, bucket, path; version="", retry=true,
-                byte_range=nothing, raw=false, headers=Dict{String, Any}(), kwargs...)
+                byte_range=nothing, raw=false,
+                headers::AbstractDict{<:AbstractString,<:Any}=Dict{String, Any}(),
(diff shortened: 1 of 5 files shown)


src/AWSS3.jl
```diff
@@ -62,22 +62,38 @@ from `path` in `bucket`.
...
+As an example of how to set custom HTTP headers, the below is equivalent to
+`s3_get(aws, bucket, path; byte_range=range)`.
 
 `s3_get(aws, bucket, path; headers=Dict{String,String}("Range" => "bytes=\$(first(range)-1)-\$(last(range)-1)"))`
 """
-function s3_get(aws::AbstractAWSConfig, bucket, path; version="", retry=true, raw=false, headers=Dict{String, Any}(), kwargs...)
+function s3_get(aws::AbstractAWSConfig, bucket, path; version="", retry=true,
+                byte_range=nothing, raw=false,
+                headers::AbstractDict{<:AbstractString,<:Any}=Dict{String, Any}(),
+                return_stream::Bool=false, kwargs...)
```
mattBrzezinski, line 79: Can we run JuliaFormatter w/ BlueStyle on this, and specify a type for `byte_range`?

src/s3path.jl
```diff
@@ -400,7 +400,9 @@ function Base.readdir(fp::S3Path; join=false, sort=true)
     end
 end
 
-Base.read(fp::S3Path) = Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key; raw=true))
+function Base.read(fp::S3Path; byte_range=nothing)
+    Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key; raw=true, byte_range))
```
mattBrzezinski, line 404:
```suggestion
    return Vector{UInt8}(s3_get(fp.config, fp.bucket, fp.key; raw=true, byte_range=byte_range))
```

src/AWSS3.jl
ExpandingMan: Personally I strongly prefer to use appropriate type assertions on function arguments in most places, but I was trying not to break to starkly with the conventions in this package.  The assertion on `headers` was added because I was afraid that strange things can happen if it was passed a different type, given the code below.

This should be resolved now.

ExpandingMan:
No rush, but I just wanted to bump this in case it was forgotten.

mattBrzezinski (approved):
bors r+

ExpandingMan:
Note that I can't see any of this output, so if there's something wrong here you're either going to have to provide a link or just tell me what it is.

mattBrzezinski:
> Note that I can't see any of this output, so if there's something wrong here you're either going to have to provide a link or just tell me what it is.

You're unable to see this page?

- https://github.com/JuliaCloud/AWSS3.jl/runs/3067917673

ExpandingMan:
Oh, yes I can see that, perhaps I clicked on the wrong thing.

Failing on an old version of Julia because of missing `Base` functions.  Will change.

ExpandingMan:
Yay, thanks!  Should be non-breaking, so a tag would be nice when you get a chance.

Issue #168, opened 2021-07-08
S3Paths: get config at use-time, not construction time?


ericphanson:
If you create an S3Path, by default it calls `global_aws_config()` to set the config. I can see why it does that, but I think a better default would be to set the config to `nothing`, and then we can replace calls to `fp.config` by something like `@something(fp.config, global_aws_config())` (that could even be a `getproperty` overload).

What is the difference? If you use S3Paths as hardcoded constants like
```julia
module MyPkg
const DATASET_PATH = S3Path("s3://...")

...
end
```
then you need s3 access at precompile time (and `using` time) otherwise the `global_aws_config()` command will throw. Also, if you do
```julia
module MyPkg
const DATASET_PATH = S3Path("s3://...")

function __init__()
    global_aws_config(; my_settings...)
end
...
end
```
then in fact `DATASET_PATH` will be constructed *before* `__init__` is called, so you will get the wrong config settings (i.e. the default ones, not the ones you set in `__init__`). (This gave @ararslan a tricky error the other day).

I think in general it's kind of a "gotcha", because users usually think of paths as a simple stringy-object-that-points-at-a-resource, not something stateful. For example, if I construct a path, realize my config is wrong, fix my global config, then I need to construct the path again-- but new users might not realize that and confused/frustrated as to why it still doesn't work.

Note: this would only apply to the default construction; if you pass a config at construction time then that config would be used always (since `fp.config` would not be nothing, it would be a config).

#### If we need to always fetch the config when using a path, would it be slow?

I don't think so:
```julia
julia> function f()
       config = global_aws_config()
       return config.region # do something with the config
       end
f (generic function with 1 method)

julia> @benchmark f()
BechmarkTools.Trial: 10000 samples with 998 evaluations.
 Range (min … max):  17.062 ns … 33.830 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     17.092 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.168 ns ±  0.832 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▃▄▇██▇▆▅▃▂▂▁   ▁▃▃▁▁                                      ▂
  ▃▇██████████████████████▆▇▇█▇▅▆▇▆▆▆▆▆▇▆▅▆▅▅▅▄▄▅▃▁▅▃▃▄▁▃▁▃▄▄ █
[...]

PR #163, opened 2021-06-18
WIP: add `minio` test script for local development

ericphanson:
cc @omus who suggested that this could be helpful.

This almost works locally:
```julia
julia --project
;cd                _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.1 (2021-04-23)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

shell> cd test
/Users/eph/AWSS3.jl/test

julia> include("minio.jl")
[ Info: Precompiling AWSS3 [1c724243-ef5b-51ab-93f4-b0a88ac62a95]
[ Info: Initializing PathSet{S3Path{MinioConfig}}
s3_folders_and_files: Error During Test at /Users/eph/AWSS3.jl/test/s3path.jl:150
  Got exception outside of a @test
  AWSException: XMinioParentIsObject -- Object-prefix is already an object, please choose a different object-prefix name.
  HTTP.ExceptionRequest.StatusError(400, "PUT", "/ocaws.jl.test.20210618t163010z/pathset-root/foobar/car.txt", HTTP.Messages.Response:
  """
  HTTP/1.1 400 Bad Request
  Accept-Ranges: bytes
  Content-Length: 452
  Content-Security-Policy: block-all-mixed-content
  Content-Type: application/xml
  Server: MinIO
  Vary: Origin
  X-Amz-Request-Id: 1689BA2A3591C6F8
  X-Xss-Protection: 1; mode=block
  Date: Fri, 18 Jun 2021 16:30:31 GMT
  
  <?xml version="1.0" encoding="UTF-8"?>
  <Error><Code>XMinioParentIsObject</Code><Message>Object-prefix is already an object, please choose a different object-prefix name.</Message><Key>pathset-root/foobar/car.txt</Key><BucketName>ocaws.jl.test.20210618t163010z</BucketName><Resource>/ocaws.jl.test.20210618t163010z/pathset-root/foobar/car.txt</Resource><RequestId>1689BA2A3591C6F8</RequestId><HostId>fbc72e52-c23e-4060-88e8-eed760c5149c</HostId></Error>""")
  
  Stacktrace:
[...]

PR #175, opened 2021-07-19
use latest config in S3Path's unless passed explicitly

ericphanson:
Closes https://github.com/JuliaCloud/AWSS3.jl/issues/168

- [x] needs tests
- [x] decide if its breaking (edit: likely should be breaking to ensure there aren't subtle errors downstream, even though it's somewhat unlikely this behavior is being relied on)

As an aside, I think this plus https://github.com/JuliaCloud/AWSS3.jl/pull/161 will make S3Paths very ergonomic for reproducible usage, e.g.

```julia
module MyModellingPackage

const MyDataSet = S3Path(bucket, key; version)

...fun modelling work...

end # module
```
Then, since it's hardcoded, bumping the dataset means making a new commit, so the code and data versions are always tied for reproducibility (assuming you commit a manifest of course).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.8.6"
 [deps]
 AWS = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 EzXML = "8f5d6c58-4d21-5cfd-889c-e3ad7ee6a615"
 FilePathsBase = "48062228-2e41-5def-b9a4-89aafe57970f"
@@ -18,6 +19,7 @@ XMLDict = "228000da-037f-5747-90a9-8195ccbf91a5"
 
 [compat]
 AWS = "1.25"
+Compat = "3.29.0"
 EzXML = "0.9, 1"
 FilePathsBase = "0.9"
 HTTP = "0.8, 0.9"
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -32,7 +32,7 @@ using EzXML
 using Dates
 using Base64
 using UUIDs
-using Compat
+using Compat: @something
 
 @service S3
 
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -56,7 +56,7 @@ using Compat: @something
 
 const SSDict = Dict{String,String}
 const AbstractS3Version = Union{AbstractString,Nothing}
-const AbstractS3PathConfig = Union{AbstractAWSConfig, Nothing}
+const AbstractS3PathConfig = Union{AbstractAWSConfig,Nothing}
 
 __init__() = FilePathsBase.register(S3Path)
 
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -56,6 +56,7 @@ using Compat: @something
 
 const SSDict = Dict{String,String}
 const AbstractS3Version = Union{AbstractString,Nothing}
+const AbstractS3PathConfig = Union{AbstractAWSConfig, Nothing}
 
 __init__() = FilePathsBase.register(S3Path)
 
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -32,6 +32,7 @@ using EzXML
 using Dates
 using Base64
 using UUIDs
+using Compat
 
 @service S3
 
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -20,8 +20,8 @@ function S3Path(
     root::AbstractString,
     drive::AbstractString,
     isdirectory::Bool,
-    config::AbstractS3PathConfig,
     version::AbstractS3Version,
+    config::AbstractS3PathConfig,
 )
     return S3Path{typeof(config)}(segments, root, drive, isdirectory, version, config)
 end
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -104,7 +104,7 @@ end
 function S3Path(
     str::AbstractString;
     version::AbstractS3Version=nothing,
-    config::AbstractS3PathConfig=global_aws_config(),
+    config::AbstractS3PathConfig=nothing,
 )
     result = tryparse(S3Path, str; config=config)
     result !== nothing || throw(ArgumentError("Invalid s3 path string: $str"))
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -117,7 +117,9 @@ function S3Path(
     return result
 end
 
-function Base.tryparse(::Type{<:S3Path}, str::AbstractString; config::AbstractS3PathConfig=nothing)
+function Base.tryparse(
+    ::Type{<:S3Path}, str::AbstractString; config::AbstractS3PathConfig=nothing
+)
     uri = URI(str)
     uri.scheme == "s3" || return nothing
 
@@ -204,7 +206,9 @@ end
 # Use `fp.config` unless it is nothing; in that case, get the latest `global_aws_config`
 get_config(fp::S3Path) = @something(fp.config, global_aws_config())
 
-FilePathsBase.exists(fp::S3Path) = s3_exists(get_config(fp), fp.bucket, fp.key; version=fp.version)
+function FilePathsBase.exists(fp::S3Path)
+    return s3_exists(get_config(fp), fp.bucket, fp.key; version=fp.version)
+end
 
 Base.isfile(fp::S3Path) = !fp.isdirectory && exists(fp)
 function Base.isdir(fp::S3Path)
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -40,7 +40,7 @@ end
 
 """
     S3Path()
-    S3Path(str; config::$(AbstractS3PathConfig)=nothing, version=nothing)
+    S3Path(str; version::$(AbstractS3Version)=nothing, config::$(AbstractS3PathConfig)=nothing)
 
 Construct a new AWS S3 path type which should be of the form
 "s3://<bucket>/prefix/to/my/object".
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -1,4 +1,6 @@
-struct S3Path{A<:Union{Nothing, AbstractAWSConfig}} <: AbstractPath
+const S3PathConfig = Union{AbstractAWSConfig, Nothing}
+
+struct S3Path{A<:S3PathConfig} <: AbstractPath
     segments::Tuple{Vararg{String}}
     root::String
     drive::String
@@ -8,13 +10,13 @@ end
 
 # constructor that converts but does not require type parameter
 function S3Path(segments, root::AbstractString, drive::AbstractString, isdirectory::Bool,
-                config::Union{Nothing, AbstractAWSConfig})
+                config::S3PathConfig)
     S3Path{typeof(config)}(segments, root, drive, isdirectory, config)
 end
 
 """
     S3Path()
-    S3Path(str; config::Union{Nothing, AbstractAWSConfig}=nothing)
+    S3Path(str; config::$(S3PathConfig)=nothing)
 
 Construct a new AWS S3 path type which should be of the form
 "s3://<bucket>/prefix/to/my/object".
@@ -48,13 +50,13 @@ function S3Path()
     )
 end
 # below definition needed by FilePathsBase
-S3Path{A}() where {A<:Union{Nothing, AbstractAWSConfig}} = S3Path()
+S3Path{A}() where {A<:S3PathConfig} = S3Path()
 
 function S3Path(
     bucket::AbstractString,
     key::AbstractString;
     isdirectory::Bool=false,
-    config::Union{Nothing, AbstractAWSConfig}=nothing,
+    config::S3PathConfig=nothing,
 )
     return S3Path(
         Tuple(filter!(!isempty, split(key, "/"))),
@@ -69,7 +71,7 @@ function S3Path(
     bucket::AbstractString,
     key::AbstractPath;
     isdirectory::Bool=false,
-    config::Union{Nothing, AbstractAWSConfig}=nothing,
+    config::S3PathConfig=nothing,
 )
     return S3Path(
         key.segments,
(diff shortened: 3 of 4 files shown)


ericphanson:
bors try

Project.toml
```diff
@@ -18,6 +19,7 @@ XMLDict = "228000da-037f-5747-90a9-8195ccbf91a5"
 
 [compat]
 AWS = "1.25"
+Compat = "3.29.0"
```
ericphanson, line 22: minimum version to get https://github.com/JuliaLang/Compat.jl/pull/742 (but this syntax means this version or any later in the v3 series)

src/AWSS3.jl
```diff
@@ -32,6 +32,7 @@ using EzXML
 using Dates
 using Base64
 using UUIDs
+using Compat
```
ericphanson, line 35: this is a dependency of AWS.jl anyway, so not a new dependency in the stack

ericphanson:
bors try

ericphanson:
No tests failed, which is a bit encouraging that it's not breaking. Basically, the case that will fail is:

```julia
global_aws_config(; arguments_1...)
path = S3Path(bucket, key) # only valid with `arguments_1`

global_aws_config(; arguments_2...)
read(path) # works on master/release; fails on this PR because it picks up the new config
```

On release/master, `path` will hold the config with `arguments_1` so this will work, whereas with this PR, it will pick up the latest global config (with `arguments_2`) and fail.

However, if you pass a config like
```julia
config = global_aws_config(; arguments_1...)
path = S3Path(bucket, key; config) # only valid with `arguments_1`

global_aws_config(; arguments_2...)
read(path) # still works
```
then it will "freeze" that config in like it does now. So it's only about the default when you don't pass a config explicitly.

I think the current behavior is surprising/unintuitive; most users expect a "path" to just be a stateless indicator/pointer/etc to an asset, not something stateful that depends on *when* it was created. Therefore, I think there's a decent argument to be made that this is a bugfix and likely anyone who wanted to create their paths with a particular config (different from the global one they'd be using later) would already be passing a config explictly. However, given AWSS3 likely has a large set of users, I also wouldn't be that surprised if someone *was* relying on freezing the global config in a path and then changing the global config.

If anyone has any thoughts about whether this should be a breaking release or not please share them. I think we shouldn't be too afraid to make breaking releases that improve the semantics, however, especially since we are pre-1.0 anyway, so I would be fine with just doing the breaking version bump here.

ericphanson:
bors try

ericphanson:
bors try

ExpandingMan:
In my opinion this should be a breaking change because the default behavior has changed.  Had you left the defaults alone, it would be non-breaking.

On second thought, it makes me a little nervous to change the default at all... it seems reasonable to expect `S3Path` not to have to suffer the overhead of making this additional request every time it is used... Not sure if anyone has any strong opinions about it, but after a few moments of reflection my instinct would be not to make this a new default.

ericphanson:
There is no additional request; once the global config has been set once it is just deferencing a ref: https://github.com/JuliaCloud/AWS.jl/blob/76619b840682cea50c4069f89e2e36ee6f63b802/src/AWS.jl#L44-L62

ExpandingMan:
Clearly I did not read this carefully enough.  That sounds great, I suppose this is non-breaking then.  Might be worth waiting for someone more lucid than I to chime in before merging.  I'm not sure if this will break anything for anyone, but I suspect that if it does it will be in a subtle and infuriating way.

ericphanson:
Ok, I think breaking is a reasonable choice then. That will help make sure no one is caught off guard by this change, and those relying on the previous behavior will have the opportuntity to update (by passing in the config as e.g. changing `S3Path(bucket, key)` to `S3Path(bucket, key; config = global_aws_config())`).

To help justify why I think this change is still worth a breaking release: the current behavior can be pretty unexpected, and if you aren't in `us-east-1` which is the default region, the wrong config means 301 errors when you try to use the path. Users often don't expect the config to be "baked in", so they are confused when they create a path (maybe at precompile time), then set the region (maybe at `__init__` time), then use the path (maybe at runtime), and then get an error. Region is just one example; any other choice requiring a custom config will be similar. Additionally, one can always pass a config object directly in order to recover the previous behavior.

rofinn:
Hmm, I wonder if this would be better handled in AWS.jl with a `LazyAWSConfig <: AbstractAWSConfig` type? Introducing that change wouldn't be  breaking to either package and would allow you to use that config outside of the `S3Path` context. Thoughts?

ericphanson:
Hm, interesting idea. Are there other objects that "bake in" the config like this? Most of what I've used AWS.jl for has been with `S3Path`s or explicit API calls like
```julia
@service CloudWatch_Logs
CloudWatch_Logs.describe_log_streams(log_group, params)
```
which gets the latest config automatically (or you can pass one with `aws_config`).

The other concern I have with that idea is that I don't think it would work to support this kind of use-case:
```julia
module MyModellingPackage

const MyDataSet = S3Path(bucket, key)

function __init__()
    global_aws_config(; region="us-east-2")
    return nothing
end

...fun modelling work...

end # module
```

Here, the `S3Path` is created at precompilation time, which means it bakes in the wrong region (since the "right" region in this scenario is set at `__init__` time). If we created a `LazyAWSConfig`, we could set it in `__init__`, but it's too late at that point, since the `S3Path` is created at `precompile` time with the wrong config.

src/s3path.jl
```diff
@@ -1,4 +1,4 @@
-struct S3Path{A<:AbstractAWSConfig} <: AbstractPath
+struct S3Path{A<:Union{Nothing, AbstractAWSConfig}} <: AbstractPath
```
omus, line 1: Shouldn't be breaking but there may be some incompatibilities with packages that dispatch on `S3Path{<:AbstractAWSConfig}`

src/s3path.jl
```diff
@@ -8,13 +8,13 @@ end
 
 # constructor that converts but does not require type parameter
 function S3Path(segments, root::AbstractString, drive::AbstractString, isdirectory::Bool,
-                config::AbstractAWSConfig)
+                config::Union{Nothing, AbstractAWSConfig})
```
omus, line 11: I'd probably define a constant for this. Not certain what to call it but this seems fine:
```
const S3PathConfig = Union{AbstractAWSConfig, Nothing}
```

src/s3path.jl
```diff
@@ -31,28 +31,30 @@ NOTES:
...
-    config = global_aws_config()
-    account_id = aws_account_number(config)
-    region = config.region
+    config = nothing
 
     return S3Path(
         (),
         "/",
-        "s3://$account_id-$region",
+        "",
```
omus, line 45: Surprising this didn't break any tests

src/AWSS3.jl
omus: It's easy to lose track of why you needed to use Compat in the first place:
```suggestion
using Compat: @something
```

omus (commented):
Overall I agree with being able to decouple an `S3Path` from a specific AWS configuration. One thought on this: is it useful to have an `S3Path` with an embedded config? I think the dynamic behaviour allowed by `nothing` here is probably what most people want. I suspect including the `config` as field was done mainly as a convenience to avoid having to pass in a configuration when making a call such as `stat(::S3Path)`.

This makes me think if we're making a breaking change anyway possibly we should remove the `config` field from `S3Path` and allow passing in an alternate config when one is actually used.

rofinn:
@omus Removing the `config` field defeats the point of the type though. At that point you'll need to know it's an S3Path at the call-site which remove the entire benefit of this type. I still think we could replace the config fields with a `Function` that returns a config. If folks want to memoize that function for efficiency they can.

omus:
I'm definitely not going to push on removing `config` as part of this PR as it'll most definitely cause this to stall out. We can discuss the idea further in the original issue: https://github.com/JuliaCloud/AWSS3.jl/issues/168

ericphanson:
bors try

ericphanson:
Bump :)

test/s3path.jl
```diff
@@ -354,4 +354,27 @@ end
...
+@testset "Global config is not frozen at construction time" begin
+    # Setup: create a file holding a string `abc`
+    path = S3Path("s3://$(bucket_name)/test_str.txt")
+    write(path, "abc")
+
+    prev_config = global_aws_config()
+    try
+        global_aws_config(region="us-east-2") # this is the wrong region!
+        path = S3Path("s3://$(bucket_name)/test_str.txt")
+        @test_throws AWS.AWSException read(path, String)
```
omus, line 367: Test is frail and will not throw an exception if the default region is set to "us-east-2".
```suggestion
    prev_config = global_aws_config()
    
    # Setup: create a file holding a string `abc`
    path = S3Path("s3://$(bucket_name)/test_str.txt")
    write(path, "abc")
    @test read(path, String) == "abc"  # Have access to read file

    alt_region = prev_config.region == "us-east-2" ? "us-east-1" : "us-east-2"
    try
        global_aws_config(region=alt_region) # this is the wrong region!
        @test_throws AWS.AWSException read(path, String)
```

omus (approved):
LGTM. Minor change to the tests to avoid a corner case

test/s3path.jl
ericphanson: good point!

ericphanson:
bors try

ericphanson:
Ok looking good! If there are no objections, tomorrow around this time I will merge and make a breaking release with the new functionality. Please chime in if you would prefer a different approach.

mattBrzezinski:
> Ok looking good! If there are no objections, tomorrow around this time I will merge and make a breaking release with the new semantics. Please chime in if you would prefer a different approach.

I would like to get a few other PRs in first so we don't need to back-port them. Specifically;

- https://github.com/JuliaCloud/AWSS3.jl/pull/163
- https://github.com/JuliaCloud/AWSS3.jl/pull/128

omus:
> I would like to get a few other PRs in first so we don't need to back-port them.

Possibly this may make it in too:
- #161

omus:
I think we're just waiting on #182 and then we can make the last 0.8 release

ericphanson:
bors try

ericphanson:
bors try

src/s3path.jl
```diff
@@ -1,4 +1,6 @@
-struct S3Path{A<:AbstractAWSConfig} <: AbstractPath
+const S3PathConfig = Union{AbstractAWSConfig, Nothing}
```
omus, line 1: Should probably be located with `AbstractS3Version` or `AbstractS3Version` should be re-located here. Adding the "Abstract" prefix to be consistent with `AbstractS3Version`
```suggestion
const AbstractS3PathConfig = Union{AbstractAWSConfig, Nothing}
```

src/s3path.jl
```diff
@@ -32,30 +34,32 @@ NOTES:
 - On top of the standard path properties (e.g., `segments`, `root`, `drive`,
   `separator`), `S3Path`s also support `bucket` and `key` properties for your
   convenience.
+- if `config` is left at its default value of `nothing`, then the
+  latest `global_aws_config()` will be used in any operations involving the
+  path. To "freeze" the config at construction time, explicitly pass an
+  `AbstractAWSConfig` to the `config` keyword argument.
```
omus, line 40:
```suggestion
- If `config` is left at its default value of `nothing`, then the
  latest `global_aws_config()` will be used in any operations involving the
  path. To "freeze" the config at construction time, explicitly pass an
  `AbstractAWSConfig` to the `config` keyword argument.
```

src/s3path.jl
```diff
@@ -97,12 +100,9 @@ function S3Path(str::AbstractString; config::AbstractAWSConfig=global_aws_config
     return result
 end
 
-# if config=nothing, will not try to talk to AWS until after string is confirmed to be an s3 path
 function Base.tryparse(::Type{<:S3Path}, str::AbstractString; config::Union{Nothing,AbstractAWSConfig}=nothing)
```
omus, line 103:
```suggestion
function Base.tryparse(::Type{<:S3Path}, str::AbstractString; config::S3PathConfig=nothing)
```

src/s3path.jl
```diff
@@ -187,7 +187,11 @@ function FilePathsBase.parents(fp::S3Path)
     end
 end
 
-FilePathsBase.exists(fp::S3Path) = s3_exists(fp.config, fp.bucket, fp.key; version=fp.version)
+# Use `fp.config` unless it is nothing; in that case, get the latest `global_aws_config`
+get_config(fp::S3Path) = @something(fp.config, global_aws_config())
```
omus, line 191: I'll mention you could use `getproperty` to make this change more seamless but I think a function is the right call personally

ericphanson:
very exciting! if your schedule allows it please feel free to pick this up, otherwise I will try to get to it soon

christopher-dG:
bors try

christopher-dG:
bors try

christopher-dG:
bors try

christopher-dG:
Tests are passing. Not sure what else needs to be done for this, if anything.

mattBrzezinski (approved):
LGTM!

ericphanson (commented):
last changes looks good to me, thanks @christopher-dG ! Not sure how we want to do merging for v0.9, there's a few breaking items on the milestone so it would be good to get them in together. Should we just call master v0.9 and start merging stuff in? (And not tag a release until the milestone is completed? We can of course still backport stuff on a branch if something comes up).

src/s3path.jl
```diff
@@ -40,7 +40,7 @@ end
 
 """
     S3Path()
-    S3Path(str; version=nothing, config::AbstractAWSConfig=global_aws_config())
+    S3Path(str; config::$(AbstractS3PathConfig)=nothing, version=nothing)
```
omus, line 43:
```suggestion
    S3Path(str; version::$(AbstractS3Version)=nothing, config::$(AbstractS3PathConfig)=nothing)
```

test/s3path.jl
```diff
@@ -547,6 +547,40 @@ function s3path_tests(config)
         end
     end
 
+    # <https://github.com/JuliaCloud/AWSS3.jl/issues/168>
+    @testset "Default `S3Path` does not hold config" begin
```
omus, line 551:
```suggestion
    @testset "Default `S3Path` does not freeze config" begin
```

omus (approved):
LGTM

omus:
> Not sure how we want to do merging for v0.9, there's a few breaking items on the milestone so it would be good to get them in together. Should we just call master v0.9 and start merging stuff in? (And not tag a release until the milestone is completed? We can of course still backport stuff on a branch if something comes up).

I would just merge the changes we want to be in v0.9 into master, bump the version, then tag a release.

PR #178, opened 2021-08-03
Draft: Optimise walkpath

### LINKED ISSUE #158: Walkpath is slow, we can make it easily 10x faster.

glennmoy:
Closes #158

I think this is what we want. I've changed the original suggestion to return a `Channel` rather than a vector to align with existing behaviour. The logic for inserting the paths into the channel when `topdown = false / true` is also bit fiddly but I hope it makes sense. I can add more comments if folks like.

We can likely improve on this even more so feedback welcome.

for comparison on the speedup on a non-trivial example
```julia

# current
julia> @time collect(walkpath(path10; topdown=true));
 24.777315 seconds (164.21 k allocations: 10.317 MiB)

julia> @time collect(walkpath(path10; topdown=false));
 28.085743 seconds (164.48 k allocations: 10.400 MiB)

# this PR
julia> @time collect(walkpath(path10; topdown=true));
  0.224652 seconds (10.23 k allocations: 563.859 KiB)

julia> @time collect(walkpath(path10; topdown=false));
  0.211930 seconds (10.31 k allocations: 572.734 KiB)
```

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -193,6 +193,37 @@ function Base.isdir(fp::S3Path)
     return iterate(objects) !== nothing
 end
 
+function FilePathsBase.walkpath(fp::S3Path; topdown=true, follow_symlinks=false, onerror=throw)
+
+    # Select objects with that prefix
+    objects = s3_list_objects(fp.config, fp.bucket, fp.key; delimiter="")
+
+    # Extract keys from the objects and remove the prefix
+    rm_key = s -> chop(s, head=length(fp.key), tail=0)
+    keys = map(rm_key, o["Key"] for o in objects)
+
+    isempty(keys) && return Channel{S3Path}()
+
+    chnl = S3Path[]
+    for k in (topdown ? keys : reverse(keys))
+        batch = S3Path[]
+
+        topdown || push!(batch, fp / k)
+        prnts = parents(Path(k))
+
+        for p in (topdown ? prnts : reverse(prnts))
+            _p = fp / p
+            !(_p in chnl) && push!(batch, _p)
+        end
+
+        topdown && push!(batch, fp / k)
+        push!(chnl, (topdown ? batch : reverse(batch))...)
+    end
+
+    result = (topdown ? chnl : reverse(chnl))
+    return Channel{S3Path{AWS.AWSConfig}}(ch -> foreach(i -> put!(ch, i), result))
+end
+
 function Base.stat(fp::S3Path)
     # Currently AWSS3 would require a s3_get_acl call to fetch
     # ownership and permission settings
@@ -349,12 +380,12 @@ function _retrieve_prefixes!(results, objects, prefix_key, chop_head)
 
     for p in objects
         prefix = _pair_or_dict_get(p, prefix_key)
-        
+
         if prefix !== nothing
             push!(results, rm_key(prefix))
         end
     end
-    
+
     return nothing
 end
 


src/s3path.jl
```diff
@@ -193,6 +193,37 @@ function Base.isdir(fp::S3Path)
...
+    # Select objects with that prefix
+    objects = s3_list_objects(fp.config, fp.bucket, fp.key; delimiter="")
+
+    # Extract keys from the objects and remove the prefix
+    rm_key = s -> chop(s, head=length(fp.key), tail=0)
+    keys = map(rm_key, o["Key"] for o in objects)
+
+    isempty(keys) && return Channel{S3Path}()
+
+    chnl = S3Path[]
```
glennmoy, line 207: I tried this originally via a task but 1) I couldn't check if `_p in chnl` while the `Channel` was still open and 2) I don't think I can call `reverse` on the `Channel` either so this seemed the easiest way to proceed even though it uses more memory.

glennmoy:
2 tests from `test_walkpath` seem to be failing. Not sure what the issue is yet.

```julia
julia> result
9-element Vector{S3Path{AWS.AWSConfig}}:
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/./"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/bar/"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/bar/qux/"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/bar/qux/quux.tar.gz"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/foo/"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/foo/baz.txt"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/fred/"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/fred/plugh"

julia> expected
7-element Vector{S3Path{AWS.AWSConfig}}:
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/bar/"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/bar/qux/"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/bar/qux/quux.tar.gz"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/foo/"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/foo/baz.txt"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/fred/"
 p"s3://ocaws.jl.test.20210803t192655z/pathset-root/fred/plugh"
```

rofinn:
Yeah, I noticed that when working on an implementation that doesn't require collecting the results. FWIW, I think #179 fixes the issue and only requires a single iteration over the `s3_list_objects` results.

```julia
# The current master branch NA... gave up after 10 min :)

# gm/walkpath
julia> @time collect(walkpath(fp; topdown=false));
 46.573418 seconds (285.25 M allocations: 12.779 GiB, 3.07% gc time)

julia> @time collect(walkpath(fp; topdown=true));
 45.402200 seconds (285.14 M allocations: 12.773 GiB, 3.40% gc time)

# rf/walkpath
julia> @time collect(walkpath(fp; topdown=false));
 17.915066 seconds (1.10 M allocations: 70.752 MiB, 0.21% gc time)

julia> @time collect(walkpath(fp; topdown=true));
 19.116504 seconds (1.10 M allocations: 70.328 MiB, 0.10% gc time)
```

src/s3path.jl
```diff
@@ -193,6 +193,37 @@ function Base.isdir(fp::S3Path)
     return iterate(objects) !== nothing
 end
 
+function FilePathsBase.walkpath(fp::S3Path; topdown=true, follow_symlinks=false, onerror=throw)
+
+    # Select objects with that prefix
```
wytbella, line 198: remind me the `prefix` here refers to which parameter?

glennmoy: It's the combination of the `fp.bucket` and `fp.key`, which is the path to the root directory you want to start searching in, e.g. `s3://mybucket/path/to/root`

PR #179, opened 2021-08-04
S3 specific walkpath

rofinn:
This is an extension of #178 which hopefully fixes the bug Glenn identified and performs all operations within a single `Channel` constructor and iteration.

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -233,23 +233,19 @@ function _walkpath!(root::S3Path, prefix::S3Path, objects, chnl; topdown=true, o
             # insert the child into the results channel
             !isempty(k) && topdown && put!(chnl, child)
 
-            # If our child is also a directory then recursively call _walkpath! to
-            # process those children.
-            recurse && _walkpath!(root, child, objects, chnl; topdown=topdown, kwargs...)
+            # Apply our recursive call for the children as necessary
+            if recurse
+                _walkpath!(
+                    root, child, objects, chnl;
+                    topdown=topdown, onerror=onerror, kwargs...
+                )
+            end
 
             # If we aren't dealing with the root and we're doing bottom up iteration then
             # insert the child ion the result channel here
             !isempty(k) && (topdown || put!(chnl, child))
         catch e
-            # If take! throws a closed channel exception then exit the loop w/o error,
-            # otherwise rethrow our error condition
-            if isa(e, InvalidStateException) && e.state === :closed
-                break
-            elseif isa(e, Base.IOError)
-                onerror(e)
-            else
-                rethrow()
-            end
+            isa(e, Base.IOError) ? onerror(e) : rethrow()
         end
     end
 end
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -193,35 +193,55 @@ function Base.isdir(fp::S3Path)
     return iterate(objects) !== nothing
 end
 
-function FilePathsBase.walkpath(fp::S3Path; topdown=true, follow_symlinks=false, onerror=throw)
-
+function FilePathsBase.walkpath(fp::S3Path; kwargs...)
     # Select objects with that prefix
     objects = s3_list_objects(fp.config, fp.bucket, fp.key; delimiter="")
 
-    # Extract keys from the objects and remove the prefix
-    rm_key = s -> chop(s, head=length(fp.key), tail=0)
-    keys = map(rm_key, o["Key"] for o in objects)
+    # Construct a new Channel using a recursive internal `_walkpath!` function
+    return Channel{typeof(fp)}(chnl -> _walkpath!(fp, fp, objects, chnl; kwargs...))
+end
 
-    isempty(keys) && return Channel{S3Path}()
+function _walkpath!(root::S3Path, prefix::S3Path, objects, chnl; topdown=true, kwargs...)
+    # Next is used for storing taken elements which belong to a different parent.
+    # Think of it like a pseudo stateful iteration.
+    next = nothing
+    while true
+        try
+            # If we have no next element saved then take! the next object
+            o = isnothing(next) ? take!(objects) : next
+            # If our key doesn't start with the prefix then we've exhausted the current
+            # prefix directory.
+            startswith(o["Key"], prefix.key) || return o
 
-    chnl = S3Path[]
-    for k in (topdown ? keys : reverse(keys))
-        batch = S3Path[]
+            # Extract the non-root part of the key
+            k = chop(o["Key"], head=length(root.key), tail=0)
 
-        topdown || push!(batch, fp / k)
-        prnts = parents(Path(k))
+            # Construct a valid child path from the key
+            child = joinpath(root, k)
 
-        for p in (topdown ? prnts : reverse(prnts))
-            _p = fp / p
-            !(_p in chnl) && push!(batch, _p)
+            # If we aren't dealing with the root and we're doing topdown iteration then
+            # insert the child into the results channel
+            !isempty(k) && topdown && put!(chnl, child)
+
+            # If our child is also a directory then recursively call _walkpath! to
+            # process those children.
+            if isdir(child)
+                next = _walkpath!(root, child, objects, chnl; topdown=topdown, kwargs...)
+            end
+
+            # If we aren't dealing with the root and we're doing bottom up iteration then
+            # insert the child ion the result channel here
+            !isempty(k) && (topdown || put!(chnl, child))
+        catch e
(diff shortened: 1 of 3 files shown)


rofinn:
bors try

rofinn:
bors try

rofinn:
bors try

rofinn:
bors try

glennmoy (commented):
Is the output of this consistent with what is expected?

previously we used to get the intermediate nodes in the graph
```julia
julia> collect(walkpath(p"s3://root/"))
p"s3://root/path/"
p"s3://root/path/to1/"
p"s3://root/path/to1/file1"
p"s3://root/path/to1/file2"
p"s3://root/path/to2/"
p"s3://root/path/to2/file2"
```

but now we only get the leaf nodes
```julia
julia> collect(walkpath(p"s3://root/"))
p"s3://root/path/to1/file1"
p"s3://root/path/to1/file2"
p"s3://root/path/to2/file2"
```

Tests pass so maybe this isn't the case or tests are incomplete?

rofinn:
> Is the output of this consistent with what is expected?

It has changed slightly to be more consistent with how S3 works. AWSS3 uses `mkdir(::S3Path)` to create "directory" objects to maintain the illusion that you're working with a normal filesystem, you can easily have buckets that don't have these objects, so that makes sense. The change here is that `walkpath` will only return the list of objects that actually exist. I can add a test and docstring to that effect if you like?

> Tests pass so maybe this isn't the case or tests are incomplete?

No, that makes sense. The tests call `mkdir` to establish the directory structure, which means this implementation would return the same results as a local call.

rofinn:
I guess one argument against the current behaviour around "directories" is that `readdir(fp)` will always print the non-existent directories, so this wouldn't be consistent with that.

src/s3path.jl
```diff
@@ -193,6 +193,59 @@ function Base.isdir(fp::S3Path)
...
+            # If we aren't dealing with the root and we're doing bottom up iteration then
+            # insert the child ion the result channel here
+            !isempty(k) && (topdown || put!(chnl, child))
+        catch e
+            # If take! throws a closed channel exception then exit the loop w/o error,
+            # otherwise rethrow our error condition
+            if isa(e, InvalidStateException) && e.state === :closed
+                break
+            else
+                rethrow()
```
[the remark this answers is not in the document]
rofinn: Actually, `onerror` is only for `IOError`s, so I've added a specific condition for that.

rofinn:
Alright, this will now toss in non-existent prefixes, in order to remain consistent with `readdir`.

rofinn:
bors try

rofinn:
bors try

rofinn:
I don't know why bors isn't completing.

rofinn:
bors try

glennmoy (approved):

test/s3path.jl
```diff
@@ -75,6 +75,32 @@ function test_s3_readpath(p::PathSet)
...
+        _bar = _root / "bar/"
+        _qux = _bar / "qux/"
+        _quux = _qux / "quux.tar.gz"
+
+        # Only write the leaf files
+        write(_baz, read(p.baz))
+        write(_quux, read(p.quux))
+
+        topdown = [_bar, _qux, _quux, _foo, _baz]
+        bottomup = [_quux, _qux, _bar, _baz, _foo]
```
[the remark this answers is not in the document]
rofinn: Correct

src/s3path.jl
```diff
@@ -193,6 +193,63 @@ function Base.isdir(fp::S3Path)
     return iterate(objects) !== nothing
 end
 
+function FilePathsBase.walkpath(fp::S3Path; kwargs...)
+    # Select objects with that prefix
+    objects = s3_list_objects(fp.config, fp.bucket, fp.key; delimiter="")
+
+    # Construct a new Channel using a recursive internal `_walkpath!` function
+    return Channel(ctype=typeof(fp)) do chnl
+        _walkpath!(fp, fp, Iterators.Stateful(objects), chnl; kwargs...)
```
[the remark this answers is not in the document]
rofinn: Yeah

src/s3path.jl
```diff
@@ -193,6 +193,63 @@ function Base.isdir(fp::S3Path)
...
+    end
+end
+
+function _walkpath!(root::S3Path, prefix::S3Path, objects, chnl; topdown=true, onerror=throw, kwargs...)
```
omus, line 206: Gripe for FilePathsBase. It would be good to have documentation on `walkpath` for its keyword arguments

src/s3path.jl
```diff
@@ -193,6 +193,63 @@ function Base.isdir(fp::S3Path)
...
+            if recurse
+                _walkpath!(
+                    root, child, objects, chnl;
+                    topdown=topdown, onerror=onerror, kwargs...
+                )
+            end
+
+            # If we aren't dealing with the root and we're doing bottom up iteration then
+            # insert the child ion the result channel here
+            !isempty(k) && (topdown || put!(chnl, child))
```
omus, line 246:
```suggestion
            !isempty(k) && !topdown && put!(chnl, child)
```

test/s3path.jl
```diff
@@ -75,6 +75,32 @@ function test_s3_readpath(p::PathSet)
     end
 end
 
+function test_s3_walkpath(p::PathSet)
+    @testset "walkpath - S3" begin
+        # Test that we still return parent prefixes even when no "directory" objects
+        # have been created by a `mkdir`, retaining consistency with `readdir`.
+        _root = p.root / "s3_walkpath/"
```
omus, line 82: We should definitely have a leaf file that exists outside of this prefix. I'm pretty sure that's already the case but maybe that should be explicitly done to ensure we aren't relying too much on behaviour defined outside of this test set.

test/s3path.jl
```diff
@@ -75,6 +75,32 @@ function test_s3_readpath(p::PathSet)
...
+
+        # Only write the leaf files
+        write(_baz, read(p.baz))
+        write(_quux, read(p.quux))
+
+        topdown = [_bar, _qux, _quux, _foo, _baz]
+        bottomup = [_quux, _qux, _bar, _baz, _foo]
+
+        @test collect(walkpath(_root; topdown=true)) == topdown
+        @test collect(walkpath(_root; topdown=false)) == bottomup
```
omus, line 98: Would be good to have a test that uses `onerror`

omus (approved):
LGTM. I didn't do a full deep dive but the testset covers mostly what I'd like to see covered

test/s3path.jl
rofinn: I'm not sure I follow, we have the pathset files that we're copying into this prefix?

src/s3path.jl
rofinn: Sure.

src/s3path.jl
rofinn: Yeah, it's a carry over from `walkdir`. I'm not entirely sure how to test those without relying on mocking though.

test/s3path.jl
rofinn: I made an issue as that change would require introducing mocking and more extensive test updates. I think a separate PR should be made to introduce mocked failure here and in several other parts of the code that does error handling.

https://github.com/JuliaCloud/AWSS3.jl/issues/180

rofinn:
Anyone know what's going on with the bors status? It seems to be stuck despite the CI passing.

mattBrzezinski:
> Anyone know what's going on with the bors status? It seems to be stuck despite the CI passing.

I believe when running bors try it won't finish because it doesn’t run the slack notify? When we r+ it’ll be all good.

test/s3path.jl
omus: We should be sure to write something outside the `_root` (e.g. `p.root / "other.txt"`) to ensure that `walkpath` doesn't doesn't traverse all prefixes in the bucket and not just those with the desired prefix.

I was also trying to mention I believe that some setup for previous tests already ensures this is the case but it's probably best to make an additional object here so that the tests in this testset are not reliant on external setup.

test/s3path.jl
omus: Thanks!

test/s3path.jl
rofinn: > We should be sure to write something outside the _root (e.g. p.root / "other.txt") to ensure that walkpath doesn't doesn't traverse all prefixes in the bucket and not just those with the desired prefix.

We already know that's the case because we're copying the leaf files from the `p.root` prefix to the `_root` sub-prefix.

Issue #105, opened 2020-10-13
Add `S3Path(path::AbstractString, isdirectory::Bool)` constructor


nickrobinson251:
I find myself writing something like
```
function f(dir::AbstractString)
    s3path = S3Path(endswith(dir, "/") ? dir : dir * "/")
    ....
end
```
in a situation where `dir` is guaranteed to be a directory, but may or may not have the trailing `/`.

Another "workaround" would be to parse out the `bucket` and `key` from `dir::AbstractString` and call
```
    s3path = S3Path(bucket, key; isdirectory=true)
```
but it seems like a `isdirectory` keyword should just be exposed in a `S3Path(str)` constructor e.g.
https://github.com/JuliaCloud/AWSS3.jl/blob/master/src/s3path.jl#L74

PR #183, opened 2021-08-10
Limit `isdir` calls in `walkpath`

rofinn:
Currently, if `s3_list_objects` return intermediate "directory" objects then we'll need to call `isdir` on the result. Depending on whether these objects include the trailing "/" this `isdir` call may query AWS. These changes attempt to eliminate that requirement by inspect the next element to determine if an object is just a prefix/directory.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -40,7 +40,7 @@ using AWS.AWSServices: s3
 using FilePathsBase
 using FilePathsBase: /, join
 using HTTP
-using OrderedCollections: OrderedDict
+using OrderedCollections: OrderedDict, LittleDict
 using SymDict
 using Retry
 using XMLDict
@@ -562,7 +562,7 @@ function s3_list_objects(
     max_items=nothing,
     kwargs...,
 )
-    return Channel() do chnl
+    return Channel(ctype=LittleDict, csize=128) do chnl
         more = true
         num_objects = 0
         token = ""
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -562,7 +562,7 @@ function s3_list_objects(
     max_items=nothing,
     kwargs...,
 )
-    return Channel(ctype=LittleDict, csize=128) do chnl
+    return Channel(; ctype=LittleDict, csize=128) do chnl
         more = true
         num_objects = 0
         token = ""
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -99,6 +99,7 @@ end
 # To avoid a breaking change.
 function S3Path(
     str::AbstractString;
+    isdirectory::Union{Bool, Nothing}=nothing,
     version::AbstractS3Version=nothing,
     config::AbstractS3PathConfig=nothing,
 )
@@ -110,7 +111,17 @@ function S3Path(
         end
         result = S3Path(result.bucket, result.key; version=version, config=result.config)
     end
-    return result
+
+    # Replace the parsed isdirectory field with an explicit passed in argument.
+    is_dir = isdirectory === nothing ? result.isdirectory : isdirectory
+
+    return S3Path(
+        result.bucket,
+        result.key;
+        isdirectory=is_dir,
+        config=result.config,
+        version=ver,
+    )
 end
 
 function Base.tryparse(
@@ -243,13 +254,13 @@ function _walkpath!(
     while true
         try
             # Start by inspecting the next element
-            next = Base.peek(objects)
+            obj = Base.peek(objects)
 
             # Early exit condition if we've exhausted the iterator or just the current prefix.
-            next === nothing && return nothing
+            obj === nothing && return nothing
 
             # Extract the non-root part of the key
-            k = chop(next["Key"]; head=length(root.key), tail=0)
+            k = chop(obj["Key"]; head=length(root.key), tail=0)
 
             fp = joinpath(root, k)
             _parents = parents(fp)
@@ -263,12 +274,19 @@ function _walkpath!(
             # is a directory too
             elseif last(_parents) == prefix
                 popfirst!(objects)
-                next_child = Base.peek(objects)
-                if next_child !== nothing && startswith(next_child["Key"], fp.key)
-                    joinpath(fp, "/")
-                else
-                    joinpath(fp, isdir(fp) ? "/" : "")
-                end
+                # If our current path is a prefix for the next path then we can assume that
+                # the current path should be a directory without needing to call `isdir`
+                next = Base.peek(objects)
+                is_dir = (next !== nothing && startswith(next["Key"], fp.key)) || isdir(fp)
+                # Reconstruct our next object and explicitly specify whether it is a
+                # directory.
+                S3Path(
+                    fp.bucket,
+                    fp.key;
+                    isdirectory=is_dir,
+                    config=fp.config,
+                    version=fp.version,
+                )
             # If our filepath is a distance descendant of the prefix then start filling in
             # the intermediate paths
             elseif prefix in _parents
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -99,14 +99,14 @@ end
 # To avoid a breaking change.
 function S3Path(
     str::AbstractString;
-    isdirectory::Union{Bool, Nothing}=nothing,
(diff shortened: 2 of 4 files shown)


rofinn:
bors try

src/AWSS3.jl
```diff
@@ -427,7 +427,7 @@ Returns an iterator of `Dict`s with keys `Key`, `LastModified`, `ETag`, `Size`,
 `Owner`, `StorageClass`.
 """
 function s3_list_objects(aws::AbstractAWSConfig, bucket, path_prefix=""; delimiter="/", max_items=nothing, kwargs...)
-    return Channel() do chnl
+    return Channel(ctype=LittleDict, csize=128) do chnl
```
rofinn, line 430: Using a sized `Channel` does seem to improve performance slightly. 128 seemed small enough to be fine?

src/s3path.jl
```diff
@@ -86,15 +86,33 @@ function S3Path(
 end
 
 # To avoid a breaking change.
-function S3Path(str::AbstractString; config::AbstractAWSConfig=global_aws_config(),
-                version::AbstractS3Version=nothing)
+function S3Path(
```
rofinn, line 89: Cleaned up this function a little as being able to write `S3Path(str; isdirectory=true)` was handy for debugging the performance issues this PR was trying to solve. If folks want I could make this a separate PR?

src/s3path.jl
```diff
@@ -210,45 +228,73 @@ end
 function FilePathsBase.walkpath(fp::S3Path; kwargs...)
     # Select objects with that prefix
     objects = s3_list_objects(fp.config, fp.bucket, fp.key; delimiter="")
+    root = joinpath(fp, "/")
 
     # Construct a new Channel using a recursive internal `_walkpath!` function
-    return Channel(ctype=typeof(fp)) do chnl
-        _walkpath!(fp, fp, Iterators.Stateful(objects), chnl; kwargs...)
+    return Channel(ctype=typeof(fp), csize=128) do chnl
```
rofinn, line 234: Again, sizing the channel had a modest improvement when scanning thousands of S3 objects.

```
julia> @benchmark collect(walkpath($fp)) seconds=60
BenchmarkTools.Trial: 13 samples with 1 evaluation.
 Range (min … max):  4.573 s …    5.472 s  ┊ GC (min … max): 0.22% … 0.19%
 Time  (median):     4.817 s               ┊ GC (median):    0.16%
 Time  (mean ± σ):   4.857 s ± 240.414 ms  ┊ GC (mean ± σ):  0.11% ± 0.10%

  ▁   ▁ ▁▁▁▁     ▁    █    ▁    █                          ▁
  █▁▁▁█▁████▁▁▁▁▁█▁▁▁▁█▁▁▁▁█▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  4.57 s         Histogram: frequency by time         5.47 s <

 Memory estimate: 25.51 MiB, allocs estimate: 469434.
 ```

size=128:
```
julia> @benchmark collect(walkpath($fp)) seconds=60
BenchmarkTools.Trial: 14 samples with 1 evaluation.
 Range (min … max):  4.197 s …    4.756 s  ┊ GC (min … max): 0.22% … 0.17%
 Time  (median):     4.429 s               ┊ GC (median):    0.19%
 Time  (mean ± σ):   4.461 s ± 171.700 ms  ┊ GC (mean ± σ):  0.11% ± 0.10%

  ▁     ▁     █   ▁  ▁▁      ▁    ▁▁             █▁        ▁
  █▁▁▁▁▁█▁▁▁▁▁█▁▁▁█▁▁██▁▁▁▁▁▁█▁▁▁▁██▁▁▁▁▁▁▁▁▁▁▁▁▁██▁▁▁▁▁▁▁▁█ ▁
  4.2 s          Histogram: frequency by time         4.76 s <

 Memory estimate: 25.37 MiB, allocs estimate: 465806.
```

File: src/s3path.jl [hunk 4]
```diff
@@ -210,45 +228,73 @@ end
...
+        _walkpath!(root, root, Iterators.Stateful(objects), chnl; kwargs...)
     end
 end
 
 function _walkpath!(root::S3Path, prefix::S3Path, objects, chnl; topdown=true, onerror=throw, kwargs...)
+    @assert root.isdirectory
```
rofinn, line 240: If folks want I could make these argument errors, but we should never hit them during normal exection. It's mostly an extra internal check to make sure any changes to this function don't break that assumption.

src/s3path.jl
```diff
@@ -210,45 +228,73 @@ end
...
             # Extract the non-root part of the key
-            k = chop(next["Key"], head=length(root.key), tail=0)
+            k = chop(obj["Key"], head=length(root.key), tail=0)
 
-            # Determine the next appropriate child path
-            # 1. Next is a direct descendant of the current prefix (ie: we have a prefix object)
-            # 2. Next is a distant descendant of the current prefix (ie: we don't have prefix objects)
             fp = joinpath(root, k)
             _parents = parents(fp)
-            child, recurse = if last(_parents) == prefix || fp.segments == prefix.segments
```
rofinn, line 238: Previously, we'd always call `isdir` and query AWS when this condition was hit.

src/s3path.jl
```diff
@@ -210,45 +228,73 @@ end
...
+
+            # If the filepath matches our prefix then pop it off and continue
+            # Cause we would have already processed it before recursing
+            child = if prefix.segments == fp.segments
```
rofinn, line 259: We should be able to skip paths that match our `prefix` as we should have added them in our parent call.

src/s3path.jl
```diff
@@ -210,45 +228,73 @@ end
...
-            else
+                continue
+            # If the filpath is a direct descendant of our prefix then check if it
+            # is a directory too
+            elseif last(_parents) == prefix
+                popfirst!(objects)
+                # If our current path is a prefix for the next path then we can assume that
+                # the current path should be a directory without needing to call `isdir`
+                next = Base.peek(objects)
+                is_dir = (next !== nothing && startswith(next["Key"], fp.key)) || isdir(fp)
```
rofinn, line 269: We only call `isdir` when our current path isn't a prefix for the next. This optimization assumes `s3_list_objects` results will continue to be returned in topdown order.

src/s3path.jl
```diff
@@ -210,45 +228,73 @@ end
...
+                # Reconstruct our next object and explicitly specify whether it is a
+                # directory.
+                S3Path(
```
rofinn, line 272: It'd be nice if there was a way to do a copy and set in Julia to just change the `isdirectory` field.

test/s3path.jl
```diff
@@ -23,6 +23,7 @@ function test_s3_constructors(ps::PathSet)
     @test S3Path(bucket_name, "pathset-root/bar/qux/"; isdirectory=true) == ps.qux
     @test S3Path(bucket_name, p"pathset-root/bar/qux"; isdirectory=true) == ps.qux
     @test S3Path(bucket_name, p"/pathset-root/bar/qux"; isdirectory=true) == ps.qux
+    @test S3Path("s3://$bucket_name/pathset-root/bar/qux"; isdirectory=true) == ps.qux
```
rofinn, line 26: This should address #105

rofinn (commented):
I can always break these up into smaller PRs, but they seemed like sufficiently small changes that getting one review would be easier?

omus (commented):
I haven't had time to do a thorough review yet. It does seem reasonable though

src/AWSS3.jl
oxinabox: sounds fine to me

src/s3path.jl
oxinabox: This seems fine to me

src/s3path.jl
oxinabox: Setfield.jl does this, no?

oxinabox (approved):

src/s3path.jl
rofinn: Cool, I didn't know about that package. I've made an issue to use that, as I think we do this in other code paths as well.

https://github.com/JuliaCloud/AWSS3.jl/issues/207

PR #190, opened 2021-08-16
Format code with JuliaFormatter

christopher-dG:
``` julia
using JuliaFormatter; format(".", BlueStyle(); verbose=true)
```

Some of these changes are IMO very ugly but 🤷


christopher-dG:
bors try

christopher-dG:
bors r+

I feel like there would perhaps be some benefit to "optimizing" some of the code to look better within the constraints of the formatter, but that can wait until another time

PR #191, opened 2021-08-16
Bump version to 0.8.8

christopher-dG:


ericphanson (approved):
Looks like we hit all the prereqs we wanted!

Screenshot 2021-08-16 at 17 00 10

christopher-dG:
bors r+

Good to go now I hope, but cancel if I'm wrong :)

PR #182, opened 2021-08-10
Support `S3Path` construction from string containing object version

hannahilea:
Follow-up from #161 (ref comment: https://github.com/JuliaCloud/AWSS3.jl/pull/161#discussion_r671592464).

PR #194, opened 2021-08-16
fix exists functions for directory on AWS S3

ExpandingMan:
This may be a discrepancy between AWS S3 and min.io behavior.  On the S3 bucket I am dealing with, `s3_exists` is returning `false` for a directory that most certainly exists.  My solution is to add an `isdir` check to `exists` (but *NOT* for `s3_exists`).  This shouldn't break anything for anyone.

Btw, tests are failing locally for me (for reasons unrelated to this PR), no clue why.  Will see what the github tests look like before I decide to do anything about it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "AWSS3"
 uuid = "1c724243-ef5b-51ab-93f4-b0a88ac62a95"
-version = "0.8.8"
+version = "0.8.9"
 
 [deps]
 AWS = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -208,7 +208,7 @@ function FilePathsBase.parents(fp::S3Path)
 end
 
 function FilePathsBase.exists(fp::S3Path)
-    return s3_exists(fp.config, fp.bucket, fp.key; version=fp.version)
+    return s3_exists(fp.config, fp.bucket, fp.key; version=fp.version) || isdir(fp)
 end
 Base.isfile(fp::S3Path) = !fp.isdirectory && exists(fp)
 function Base.isdir(fp::S3Path)


ericphanson:
bors try

ExpandingMan:
I'm confused... am I nuts or do these look like they have nothing to do with the change I made?

mattBrzezinski:
> I'm confused... am I nuts or do these look like they have nothing to do with the change I made?

I'll look at the AWS Account. There could be left over artifacts from previous runs somehow.

ExpandingMan:
Thanks.  Also, some of these are definitely from the change I made.  Looks like a bunch of stuff improperly uses `exists` when it should be using `isfile`.  Fix incoming.

Part of me is starting to question the wisdom of going against the design philosophy of S3 as a pure key-value store and doing this kind of directory checking, but if we want this to work like a real file path with `FilePathsBase` I guess there really isn't any choice.

ExpandingMan:
Uh... also some of what I'm seeing here is making me rather nervous... some of this code is most definitely assuming that `exists` works on directories.  For me it most definitely did not work.

Can anyone else confirm whether or not `s3_exists` works for directories in AWS S3?  Is it possible we are running into some kind of AWS versioning issue here?

ExpandingMan:
I also just checked Minio.jl to see if I was seeing any of those weird token issues and no, I am not, at least not on the unit tests.

mattBrzezinski:
> Uh... also some of what I'm seeing here is making me rather nervous... some of this code is most definitely assuming that `exists` works on directories. For me it most definitely did not work.
> 
> Can anyone else confirm whether or not `s3_exists` works for directories in AWS S3? Is it possible we are running into some kind of AWS versioning issue here?

```julia
julia> s3_exists("mattbr-test-bucket", "directory/")
true
```

ExpandingMan:
Ok, still have to figure out what in the world is going on with tests, but...

The existing form of `s3_exists` was a rather scary error catch.  This doesn't explain why we are seeing different behavior from AWS, at this point I can only imagine that is a version difference.  I've tried everything I can think of including messing with the stupid `/` but as near as I can tell our AWS's are just behaving differently.

I've implemented an alternate version of `s3_exists` based on `s3_list_objects`.  The way this works is that, if you call `s3_list_objects` on an object that exists exactly, you should get exactly one return value with the exact key that was given.  If you get anything else, it doesn't exist.  This might be slower in some circumstances but is likely faster in others.  Obviously I'll have to figure out what disaster has happened with the unit tests before I can prove this works.

src/AWSS3.jl
```diff
@@ -202,19 +202,11 @@ Is there an object in `bucket` at `path`?
...
-
-    catch e
-        @delay_retry if ecode(e) in ["NoSuchBucket", "404", "NoSuchKey", "AccessDenied"]
-        end
-
-        @ignore if ecode(e) in ["404", "NoSuchKey", "AccessDenied"]
-            return false
-        end
-    end
+    l = s3_list_objects(aws, bucket, path; kwargs...)
```
rofinn, line 205: Isn't this a more expensive operation though? Maybe at least set `max_items=1`?

src/s3path.jl
```diff
@@ -298,7 +305,8 @@ function Base.stat(fp::S3Path)
     s = 0
     last_modified = DateTime(0)
 
-    if exists(fp)
+    # this only works for files and not directories
+    if isfile(fp)
```
rofinn, line 309: To clarify, this is because you can only call `s3_get_meta` on objects? So a "directory" object that ends with "/" could also work here?

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "AWSS3"
 uuid = "1c724243-ef5b-51ab-93f4-b0a88ac62a95"
-version = "0.8.8"
+version = "0.8.9"
```
rofinn, line 3: This seems like a breaking change?

rofinn (commented):
Seems mostly fine, but this may have some performance issues along with technically changing the existing behaviour.

ExpandingMan:
I got a bunch of things wrong here.  This may not work.  It is *very* annoying that AWS doesn't have a straightforward way of checking if something exists, all the alternatives are bad.

Still messing with this...

src/AWSS3.jl
ExpandingMan: I've changed this to `max_items=1`.  Still haven't done anything like a thorough analysis of how the performance is.

src/s3path.jl
ExpandingMan: It's pretty cagey... there is really no such thing as a "directory" in S3, but this package pretends like there is to emulate an OS filesystem interface.  If you call `mkdir` with this package, it will create an object for you, but as far as S3 is concerned it's an object like any other.  In those cases, technically this function would work.  However, if you e.g. create `dir/file`, then this will fail for `dir/`.  I'm sure there is more to be done here, but for now, this seems like a workable solution.

ExpandingMan:
Btw, please see [this issue](https://github.com/JuliaCloud/AWSS3.jl/issues/196).

ExpandingMan:
Ok, I've been chasing my tail a bit.  The situation is quite confusing.  I think the core of the issue is that S3 doesn't really have a concept of directories but some features have been hacked in to emulate them.

Anyway, `s3_list_objects` version of `s3_exists` does not seem to work because it returns the same value whether the file just doesn't exist, or a blank node was created i.e. with `mkdir`.  There might be a bigger fix here also involving `s3_list_objects` but I'm still sorting it all out, may not do it until tomorrow now.

ExpandingMan:
Funny how these things start as "oh, I'll just fix that one little thing, it'll take 10 minutes" and it turns into a major ordeal resulting in significant breaking changes.

Ok, I've done unexpectedly much here, but I think the result is good.  There are a number of things to explain.

First, what's the deal with empty "directories"?  Well, once upon a time, the people at AWS designed key-value store with no concept of directories, but this presumably broke all sorts of things that people were trying to do that treat it like a file system.  The inevitable result is that there are now some hacks in the system that make empty files mostly but not quite work like empty directories.  These *ARE* supported by a number of features: in particular, if you look at the little GUI file browser in min.io these show up as directories, but more importantly, this is the *only* type of file that will get ignored if you try to create another file "inside" it, that is if you create `somefile/` you can also create `somefile/someotherfile` iff `somefile/` is empty.  However, other properties of empty files get kind of weird.  In particular, they only show up in `list_objects` if you really know what you're doing (but yes, you can see them, it's very confusingly inconsistent).  The trouble with all this is that you have to really know what you're doing to check if an "empty directory" exists.  Furthermore, if you want to check that just any ol' directory exists, you have to be very careful because "directories" and "empty directories" are *completely different objects*: "directories" simply appear in the names of more deeply nested keys, but "empty directories" are their own nodes (again, only if you know how to look for them).

Sigh, tired yet? I know I am.

Anyway, here is the list of things that I have changed:
- `s3_list_objects` now behaves in a way that makes *WAY* more sense, and it is based on `list_objects_v2` and it supports `start-after`.  Previously, this function would return *at least* the path queried regardless of whether it exists, because it was adding from a field in the return dict that was simply echoing the input argument.  This of course makes no sense, so the result is a vast improvement, and I tried to keep it roughly in line with what appears to be the intended use of `list_objects`.  This is of course a fairly significant breaking change (though no unit tests needed to be changed specifically for this, so maybe not), but I would expect everyone to agree that this behavior is vastly preferable to the old one, so best get it in now.
- `s3_exists` is now base on `s3_list_objects_v2` and it is very careful to work in all cases: files, directories, empty directories.  To do this, it uses a "trick" whereby, if you are searching for `dir/`, it looks for `dir.` using `start-after`.  The reason for this peculiar choice is that `.` happens to be the character in the ASCII table immediately preceding `/`.  Huh, what?  Well, you see, checking for `dir/` directly will not turn up anything if it's empty, regardless of whether it exists. Yikes.  Btw, I have done some very rough benchmarks, and this doesn't seem much slower than using `s3_get_meta`, and I believe it will be significantly faster when responding in the negative because it doesn't involve any error catching.
- Some other functions such as `isdir` and `mkdir` required some changes to accommodate the aforementioned.
- I added a `test/sandbox.jl` script to make it much easier for anyone wanting to work on this package to set up a little min.io bucket to test all the stuff I'm talking about here.  I think it'd be nice to keep this in the repo, I know that I for one will just wind up reaching for it next time I work on this package.
- I've done a number of min.io related fixes for more recent versions.  Most importantly, by default it doesn't like using multiple directories as arguments, it seems to expect each provided directory to be running on physically separate hardware from the others.  I gave up on `s3_enable_versioning` for minio, there might still be a way to do it but I don't know how.
- By now these are *definitely* breaking changes, so I bumped to `0.9`.  Minio.jl will require a compat bump, I've just committed that and will tag it tomorrow.

There's probably more I need to say here, but I can't remember by now so... let's see how tests do!  Everything is passing on min.io for me, so we'll see.

ericphanson:
I haven’t gotten a chance to look closely at this but consistency improvements around empty directories sound good to me :). We are planning on making a breaking release for #175 soon anyway so it seems like a good time for that.

bors try

ExpandingMan:
This failure is because the registered version of Minio.jl cannot support AWSS3.jl `0.9`.  I will tag it today and that should take care of it.  Will post back here when it's done.

Project.toml
ExpandingMan: After the rest of my changes, yes this is unambiguously breaking.

ericphanson:
We can also just not bump the version on this PR at all-- I'll tag it on the v0.9 milestone so it's marked that way

ExpandingMan:
Ok, I've reduced the version for now so we can get tests running sooner rather than later.

ericphanson:
bors try

src/AWSS3.jl
```diff
@@ -194,27 +194,33 @@ end
...
+    c = get(l, "Contents", nothing)
+    isnothing(c) && return false
+    return get(c, "Key", "") == path
+end
+
+function _s3_exists_dir(aws::AbstractAWSConfig, bucket, path; kw...)
+    a = string(path)[1:end-1] * "."
+    # note that you are not allowed to use *both* `prefix` and `start-after`
+    q = Dict("delimiter" => "", "max-keys" => 1, "start-after" => a)
+    l = S3.list_objects_v2(bucket, q; aws_config=aws)
```
ericphanson, line 209: will this break down if there actually is a path with a `.`? since `max-keys=1` means we might just get that one and not the next one, right?

(not that I totally get what's going on here...)

ericphanson (commented):
these changes look mostly reasonable to me but it would help a lot to have a bunch of tests with the various edge cases you're sorting out here, so we can see it's really doing the right thing now (and to not regress, since this is obviously tricky stuff)

ExpandingMan:
Yeah, I should add more tests for sure.

Also, I may have inadvertently removed some functionality around path versioning... a little worrying as now I'll have to figure out if it's even possible to support the way I now have `exists` set up (which I promise is way better).

src/AWSS3.jl
ExpandingMan: I was careful to check that case, and as far as I can tell, no, this does not break.  I will add some unit tests.

ExpandingMan: (And yeah, it's horrendously confusing, I'll probably have to document it better.)

ExpandingMan: Oh, I see what you were saying... yeah, its' very tricky, so `start-after` means start *AFTER*, that means this does *NOT* pick up `dir.` the first thing it does pick up is `dir/`.  That's why this should pick up that object and only that object iff it exists.  yes, super confusing

ericphanson: ohh ok, got it!

ExpandingMan:
Alright, just want to update where I am since I'm a little worried this will get stalled out now...

Versioning stuff is pretty scary.  Support for it in the API is *extremely* half-assed.  Most functions do not support any kind of version specification argument, so to use this feature you are stuck relying on some very limited functionality.

I think probably the best thing to do here is to continue with the approach I already implemented but to have some fallbacks in cases where the version is specified, particularly in `S3Path`, but testing this is going to be scary because, as far as I can tell, you need multiple physical min.io drives to use versioning.

If anyone knows of an easy way to enable versioning on a local min.io server it would be much appreciated.

I also would like to hear opinions on how much we want AWSS3.jl's API to support versioning when the AWS API is working so hard against it.  Obviously we need `S3Path` to just do exactly what it's been doing, anything else would be too unacceptably breaking to anybody actually using versioning, but I'm less sure about the `s3_` functions.  For example, do we want a separate `s3_version_exists`?  This might seem strictly worse than what's already there at first glance, but like I said, the AWS API support for versioning *sucks*, so separating it out a bit might make our own API more flexible.

ExpandingMan:
Ok, everything ought to be back to working and I've added a bunch of tests for edge cases surrounding my improvements.

A summary:
- I have added the new functions `s3_exists_versioned` and `s3_exists_unversioned`.  The former is the same as `s3_exists` before this PR: like I said, the AWS API support for versioning is abysmal, there are really no good options her.  There is surely *some* way of doing this without error catching, but at this point life is just too short.
- The function `s3_exists` now accepts the same arguments as before this PR.  It uses `s3_exists_unversioned` by default, but if a non-null version is passed it will use `s3_exists_versioned`.  Frankly, I'd prefer to remove `s3_exists_unversioned` and just make it the default behavior for `s3_exists`, and if you want versioning you can use the other function, my justification for this being that if AWS wants to do a shit job supporting versioning, why should we try so hard to include it, but I can see how many would feel that is an overly gratuitous breaking change, so I don't want to die on that hill.
- A few more docs clarifying some of what I've explained here, and preparing for an upcoming docs overhaul PR.

Ok, let's hope it passes this time.  Having to wait to see if it actually passes on AWS is always kind of scary :fearful:

ericphanson:
bors try

ericphanson:
> The min.io docs specifically state that they only support versioning in a distributed context, so it sounds like this is pretty dead in the water, at least on recent versions.

In #163 I tried by adding several tmp directories (so maybe it would think it was distributed); that at least let me enable versioning, but when I tried to do anything with it, it would give signature mismatch errors and such.

> The function s3_exists now accepts the same arguments as before this PR. It uses s3_exists_unversioned by default, but if a non-null version is passed it will use s3_exists_versioned. Frankly, I'd prefer to remove s3_exists_unversioned and just make it the default behavior for s3_exists, and if you want versioning you can use the other function, my justification for this being that if AWS wants to do a shit job supporting versioning, why should we try so hard to include it, but I can see how many would feel that is an overly gratuitous breaking change, so I don't want to die on that hill.

Well we at Beacon use versioning and recently contributed some functionality for it with S3Paths (#155, #182) so I'd definitely not like to remove versioning functionality. But having a fast path for the unversioned case sounds great to me. Agreed the try-catch solution is not very nice but I guess it's what we have at this point.

> A few more docs clarifying some of what I've explained here, and preparing for an upcoming docs overhaul PR.

Awesome, thanks!

src/AWSS3.jl
```diff
@@ -13,6 +13,8 @@ export S3Path,
     s3_put,
     s3_get,
     s3_get_file,
+    s3_exists_versioned,
+    s3_exists_unversioned,
```
ericphanson, line 17: I think it might be good to keep these internal as implementation details behind `s3_exists`

src/AWSS3.jl
```diff
@@ -493,24 +541,24 @@ function s3_list_objects(
...
-            end
-            if marker != ""
-                q["marker"] = marker
+            for (v, name) in [
+                (path_prefix, "prefix"),
+                (delimiter, "delimiter"),
+                (start_after, "start-after"),
+                (token, "continuation-token"),
+            ]
+                isempty(v) || (q[name] = v)
```
ericphanson, line 561: `(key, value)` order makes it much easier to parse for me:
```suggestion
            for (name, v) in [
                ("prefix", path_prefix),
                ("delimiter", delimiter),
                ("start-after", start_after),
                ("continuation-token", token),
            ]
                isempty(v) || (q[name] = v)
```

src/s3path.jl
```diff
@@ -307,7 +313,9 @@ function Base.mkdir(fp::S3Path; recursive=false, exist_ok=false)
     else
         if hasparent(fp) && !exists(parent(fp))
             if recursive
-                mkdir(parent(fp); recursive=recursive, exist_ok=exist_ok)
+                # don't try to create buckets this way, minio at least really doesn't like it
```
ericphanson, line 316: yeah, I think it's totally fine to not allow `mkdir` to make a bucket, that would be surprising to me if it happened. It's not like `mkdir` makes a new drive on windows or something if you do `mkdir("E://")`.

test/awss3.jl
```diff
@@ -205,4 +205,25 @@ function awss3_tests(config)
             @test_throws AWS.AWSException s3_delete_bucket(config, bucket_name)
         end
     end
+
+    # these tests are needed because lack of functionality of the underlying AWS API makes certain
+    # seemingly inane tasks incredibly tricky: for example checking if an "object" (file or
+    # directory) exists is very subtle
+    @testset "path naming edge cases" begin
+        s3_create_bucket("name-tests")
```
ericphanson, line 213: do we need a new bucket here? we might need to think about naming, the bucket we've been using it
```julia
bucket_name = "ocaws.jl.test." * lowercase(Dates.format(now(Dates.UTC), "yyyymmddTHHMMSSZ"))
```
so we probably should coordinate with @mattBrzezinski about names if we want to use a new one. (Maybe I should not have been running CI on this PR actually because of this-- sorry!)

test/awss3.jl
```diff
@@ -205,4 +205,25 @@ function awss3_tests(config)
...
+        @test s3_exists("name-tests", "testdir/")
+        @test s3_exists("name-tests", "testdir.")
+        @test !s3_exists("name-tests", "testdir")
+        # the below is just to make sure this thing doesn't fail
+        @test_nowarn s3_put("name-tests", "testdir/testfile.txt", "what up")
+        @test s3_exists("name-tests", "testdir/testfile.txt")
+        # make sure the directory still "exists" even though there's a key in there now
+        @test s3_exists("name-tests", "testdir/")
+        # but it is still a directory and not an object
+        @test !s3_exists("name-tests", "testdir")
```
ericphanson, line 227: I added some `isfile` and `isdir` tests (which are possibly wrong?):
```suggestion
        s3_put("name-tests", "testdir.", "") # create an empty file called `testdir.`
        s3_put("name-tests", "testdir/", "") # create an empty file called `testdir/` which AWS will treat as an "empty directory"
        @test s3_exists("name-tests", "testdir/")
        @test isdir(S3Path("name-tests", "testdir/"))
        @test !isfile(S3Path("name-tests", "testdir/"))
        @test s3_exists("name-tests", "testdir.")
        @test isfile(S3Path("name-tests", "testdir."))
        @test !isdir(S3Path("name-tests", "testdir."))
        @test !s3_exists("name-tests", "testdir")
        
        s3_put("name-tests", "testdir/testfile.txt", "what up")
        @test s3_exists("name-tests", "testdir/testfile.txt")
        @test isfile(S3Path("name-tests", "testdir/testfile.txt"))
        # make sure the directory still "exists" even though there's a key in there now
        @test s3_exists("name-tests", "testdir/")
        @test isdir(S3Path("name-tests", "testdir/"))
        @test !isfile(S3Path("name-tests", "testdir/"))

        # but it is still a directory and not an object
        @test !s3_exists("name-tests", "testdir")
```

test/runtests.jl
```diff
@@ -27,13 +27,12 @@ include("awss3.jl") # creates `s3path_tests(config)`
         # We then run all tests with s3 directly.
 
         # We use multiple directories so that Minio can support versioning.
-        root = mktempdir()
-        dirs = [mkdir(joinpath(root, string(i))) for i in 1:12]
         port = 9005
-        minio_server = Minio.Server(dirs; address="localhost:$port")
+        minio_server = Minio.Server([mktempdir()]; address="localhost:$port")
```
ericphanson, line 31: maybe this change is why you need to disable `s3_enable_versioning` on Minio?

test/sandbox.jl
```diff
@@ -0,0 +1,45 @@
+using AWSS3, Minio, FilePathsBase
```
ericphanson, line 1: I don't really get the advantage of this as a separate script vs the minio setup in `runtests.jl`

ericphanson (commented):
thanks for digging into all this empty files stuff. I think it would be great to increase our consistency here since it's a very subtle area.

ericphanson:
By the way, in the first post you wrote

> This may be a discrepancy between AWS S3 and min.io behavior. On the S3 bucket I am dealing with, s3_exists is returning false for a directory that most certainly exists. My solution is to add an isdir check to exists (but NOT for s3_exists). This shouldn't break anything for anyone. Note that exists being broken also breaks FilePathsBase.sync for s3 to local.

It would be great to have a MWE for `s3_exists` returning false when it shouldn't to add to the tests here. I am not totally sure which of the tests you added here would fail on master.

test/awss3.jl
ExpandingMan: Ohhhh... sorry sorry, I forgot you are testing these on AWS, where this sort of thing matters more.  Apologies!  I only created the bucket to absolutely guarantee that there are no existing names in the bucket.  Regardless will fix!

test/sandbox.jl
ExpandingMan: It's more about creating the bucket and a few names.  :shrug:   Up to you of course, but every time I needed to check something I needed this, figured I'd leave it around as a courtesy to the next person.  If you want it out it's out.

test/runtests.jl
ExpandingMan: For whatever reason, this only seems to work on fairly old versions of min.io.  In more recent versions they are expecting these directories to be on physically distinct drives, and if not, it will crash when trying to start the server.  I tried getting this to work but could not.

Suggest we leave it like I now have it, and if you want to try fixing it after that'd be great, but I think it'd be best for everyone if you ensure it works on latest tagged min.io.

ExpandingMan:
> In #163 I tried by adding several tmp directories (so maybe it would think it was distributed); that at least let me enable versioning, but when I tried to do anything with it, it would give signature mismatch errors and such.

To reiterate what I said in the in-line comment, I absolutely could not get this to work on latest tagged min.io.  The documentation seems to hint that I should expect this to be the case.  Unless you have a specific suggestion and I can get it fixed, I think we should leave it.

> Well we at Beacon use versioning and recently contributed some functionality for it with S3Paths (#155, #182) so I'd definitely not like to remove versioning functionality. But having a fast path for the unversioned case sounds great to me. Agreed the try-catch solution is not very nice but I guess it's what we have at this point.

Yeah, figured it would be tough to take out.  I will add some warning in the docs.


Uh oh, the version tests are still failing... if anyone has any ideas about why, it would be hugely appreciated.  This is going to be rough for me to fix since I can't get versioning working locally...

src/AWSS3.jl
```diff
@@ -217,7 +236,36 @@ function s3_exists(
...
+
+Returns a boolean whether an object exists at  `path` in `bucket`.
+"""
+function s3_exists(
+    aws::AbstractAWSConfig, bucket, path; version::AbstractS3Version=nothing, kw...
+)
+    if version ≡ nothing
+        s3_exists_unversioned(aws, bucket, path; version=version, kw...)
+    else
+        s3_exists_versioned(aws, bucket, path; version=verion, kw...)
```
ericphanson, line 266: https://github.com/JuliaCloud/AWSS3.jl/runs/3360798119#step:7:221
```suggestion
        s3_exists_versioned(aws, bucket, path; version=version, kw...)
```

ExpandingMan: :picard-facepalm:

test/sandbox.jl
ericphanson: hm, I'll leave it up to @mattBrzezinski or @omus. If we do keep it we should include it in the tests just to make sure it doesn't error/bitrot.

ExpandingMan: Woah, woah, if you want to include it in tests than *I* definitely do not want to keep it :laughing:

test/runtests.jl
ericphanson: Ok, makes sense

ExpandingMan:
Ok, changes in this round:

- I'm praying that the reason the version tests were failing on AWS was because of my typo... like I said, it's going to be very hard for me to fix those issues so it's a little scary.
- `S3Path` constructor now sets `isdirectory` based on whether the key string ends with a `/`.  Previously, some of your new tests failed because you did not set `isdirectory`.  I felt it was better to add this as a (technically breaking) feature rather than changing the tests.
- I no longer create my own bucket, so you should be able to run this with confidence in your AWS account.  Again, apologies, I wasn't thinking when I wrote that.
- I removed `test/sandbox.jl` as it was becoming a whole thing.

Please give it a shot again on bors.

src/s3path.jl
ExpandingMan: For what it's worth, previously it just seemed to be downright broken in certain situations.  And yeah, creating the bucket with `makedir` seemed a little excessive.  Even more so since we've by now discussed extensively, `mkdir` and the functionality it makes use of on the S3 side are a bit of a hack.

ericphanson:
bors try

ExpandingMan:
Oops, that failed because the bucket gets deleted on AWS but not min.io.  Should be fixed now...

ericphanson:
bors try

src/AWSS3.jl
```diff
@@ -195,19 +195,36 @@ end
...
+function _s3_exists_file(aws::AbstractAWSConfig, bucket, path; kw...)
+    q = Dict("prefix" => path, "delimiter" => "", "max-keys" => 1)
+    l = S3.list_objects_v2(bucket, q; aws_config=aws)
+    c = get(l, "Contents", nothing)
+    isnothing(c) && return false
+    return get(c, "Key", "") == path
+end
+
+function _s3_exists_dir(aws::AbstractAWSConfig, bucket, path; kw...)
+    a = string(path)[1:(end - 1)] * "."
```
ericphanson, line 207: safer in case someone tries to make a unicode s3 path for some reason:
```suggestion
    a = chop(string(path))* "."
```

src/AWSS3.jl
```diff
@@ -217,7 +234,36 @@ function s3_exists(
         end
     end
 end
+s3_exists_versioned(a...; kw...) = s3_exists_unversioned(global_aws_config(), a...; kw...)
```
ericphanson, line 237:
```suggestion
s3_exists_versioned(a...; kw...) = s3_exists_versioned(global_aws_config(), a...; kw...)
```

ericphanson: actually if these are internal, we can just delete these methods and require a config is always passed

src/AWSS3.jl
```diff
@@ -217,7 +234,36 @@ function s3_exists(
...
+Returns a boolean whether an object exists at  `path` in `bucket`.
+
+See [`s3_exists_versioned`](@ref) to check for specific versions.
+"""
+function s3_exists_unversioned(aws::AbstractAWSConfig, bucket, path; kw...)
+    return (endswith(path, "/") ? _s3_exists_dir : _s3_exists_file)(
+        aws, bucket, path; kw...
+    )
+end
+s3_exists_unversioned(a...; kw...) = s3_exists_unversioned(global_aws_config(), a...; kw...)
```
ericphanson, line 251:
```suggestion
```

src/AWSS3.jl
```diff
@@ -217,7 +234,36 @@ function s3_exists(
...
+"""
+    s3_exists([::AbstractAWSConfig], bucket, path; version=nothing, kwargs...)
+
+Returns a boolean whether an object exists at  `path` in `bucket`.
+"""
+function s3_exists(
+    aws::AbstractAWSConfig, bucket, path; version::AbstractS3Version=nothing, kw...
+)
+    if version ≡ nothing
+        s3_exists_unversioned(aws, bucket, path; version=version, kw...)
```
ericphanson, line 262: Maybe?
```suggestion
        s3_exists_unversioned(aws, bucket, path;, kw...)
```
Also what are the keywords we're passing on? I don't see any used anywhere so maybe we should just remove them

src/AWSS3.jl
ExpandingMan: Good catch, I always think in the back of my mind that string indexing is by character...

ExpandingMan:
Good catches of my stupid mistakes, thanks!  Ok, I guess we can try yet again?

ericphanson:
bors try

ExpandingMan:
I swear there must be some kind of psychological mechanism that causes me to be more likely to break stuff that I will not be able to test locally than the stuff I will be able to test...

Ok, once more unto the breach...

ericphanson:
bors try

ericphanson:
No idea what to make of those 400s :(

ExpandingMan:
I had passed arguments to `s3_path_unversioned` incorrectly.  In my opinion this package should contain quite a bit more type assertions on arguments to prevent this sort of thing and make use cases a little clearer.

Should be fixed, please try again.

ericphanson:
bors try

ericphanson:
Getting close! Now the classic Julia 1.0 doesn’t have `isnothing`

ExpandingMan:
Fixed, please try again.

ericphanson:
bors try

ExpandingMan:
Huh?  This must be an AWS issue right?  It succeeds on mac, so it can't be a real issue can it?

ericphanson:
https://github.com/JuliaCloud/AWSS3.jl/runs/3365888163#step:7:113

Yeah, seems pretty weird, maybe a network issue or something..

let’s try again

bors try

ExpandingMan:
Wow, finally.

Please merge but hold off on tagging.  I am going to tag Minio.jl to support 0.9 and do a docs PR.

src/AWSS3.jl
```diff
@@ -195,19 +195,38 @@ end
...
+function _s3_exists_file(aws::AbstractAWSConfig, bucket, path; kw...)
+    q = Dict("prefix" => path, "delimiter" => "", "max-keys" => 1)
+    l = S3.list_objects_v2(bucket, q; aws_config=aws)
+    c = get(l, "Contents", nothing)
+    c ≡ nothing && return false
+    return get(c, "Key", "") == path
+end
+
+function _s3_exists_dir(aws::AbstractAWSConfig, bucket, path; kw...)
+    a = chop(string(path)) * "."
```
ericphanson, line 207: Can you transfer some of your comments from the PR to code comments here, to explain what’s going on with the `.`?

src/AWSS3.jl
```diff
@@ -218,6 +237,36 @@ function s3_exists(
...
+See [`s3_exists_versioned`](@ref) to check for specific versions.
+"""
+function s3_exists_unversioned(aws::AbstractAWSConfig, bucket, path; kw...)
+    return (endswith(path, "/") ? _s3_exists_dir : _s3_exists_file)(
+        aws, bucket, path; kw...
+    )
+end
+
+"""
+    s3_exists([::AbstractAWSConfig], bucket, path; version=nothing, kwargs...)
```
ericphanson, line 254: It seems the `kwargs` only do anything in the versioned case in which case they get passed to `s3_get_meta` where it gets forwarded to `head_object` where they are apparently not accepted: https://github.com/JuliaCloud/AWS.jl/blob/56febe52eb7754df869658dbaefd9519efc1aa13/src/services/s3.jl#L2616

Let’s just remove all these `kwargs…` from these exist calls, since it seems impossible to use them for anything.

ericphanson (commented):
I think it looks pretty good but I’d love to hear from other AWSS3 maintainers who might have more experience with these things before merging.

That said, no tests broke from these changes and the new tests look reasonable and are passing, so if we don’t hear from anyone in a day or two I can try to give it another careful look and then merge if it still looks good.

ExpandingMan:
Ok, I've added extensive commentary around the directory existence check and got rid of all of the extraneous `kwargs`.  I have also bumped the version since Minio 0.1.2 has now been tagged, so test should  be able to pass for this as 0.9.

ExpandingMan:
Just want to bump this as it's been languishing for a little while now.

src/AWSS3.jl
```diff
@@ -195,19 +195,58 @@ end
...
+What this function does is, given a directory name `dir/`, check for all keys which
+are lexographically greater than `dir.`.  The reason for this is that, if `dir/`
+is a 0-byte node, checking for it directly will not reveal its existence due to
+some rather peculiar design choices on the part of the S3 developers.
+
+In all such cases, if the directory exists it will be seen in the *first* item
+returned from `S3.list_objects_v2`: for empty directories this is because using
+`start-after` explicitly excludes `dir.` and `dir/` is next, for directories
+with actual keys, it is guaranteed that the first contained key will start with
+the directory name.
```
ericphanson, line 224: maybe this is slightly clearer:
```suggestion
returned from `S3.list_objects_v2`: for empty directories this is because using
`start-after` explicitly excludes `dir.` itself and `dir/` is next; for directories
with actual keys, it is guaranteed that the first contained key will start with
the directory name.
```

src/AWSS3.jl
```diff
@@ -195,19 +195,58 @@ end
...
+    c ≡ nothing && return false
+    return startswith(get(c, "Key", ""), path)
+end
+
+"""
+    s3_exists_versioned([::AbstractAWSConfig], bucket, path, version)
+
+Check if the version specified by `version` of the object in bucket `bucket` exists at `path`.
+
+Note that this function relies on error catching and may be less performant than [`s3_exists`](@ref)
```
ericphanson, line 241:
```suggestion
Note that this function relies on error catching and may be less performant than [`s3_exists_unversioned `](@ref)
```

test/awss3.jl
```diff
@@ -156,7 +156,7 @@ function awss3_tests(config)
         # https://github.com/samoconnor/AWSS3.jl/issues/24
         ctype(key) = s3_get_meta(bucket_name, key)["Content-Type"]
 
-        for k in ["file.foo", "file", "file_html", "file/html", "foobar.html/file.htm"]
+        for k in ["file.foo", "file", "file_html", "file.d/html", "foobar.html/file.htm"]
```
ericphanson, line 159: what is this change for?

test/runtests.jl
```diff
@@ -27,13 +27,12 @@ include("awss3.jl") # creates `s3path_tests(config)`
         # We then run all tests with s3 directly.
 
         # We use multiple directories so that Minio can support versioning.
```
ericphanson, line 29: should rm this comment too if we are not using multiple directories anymore:
```suggestion
```

ericphanson (approved):
looks good to me (modulo some very small comments). I'm not totally confident that we now treat the concept of a "directory" on AWS in a perfectly coherent way, but I am pretty sure this is a good improvement towards that goal. I think also using `list_files_v2` instead of checking if `s3_meta` throws in more cases is a good improvement.

If there are no objections or further review comments that need addressing, I'll plan on merging this time tomorrow.

test/awss3.jl
ExpandingMan: Because `file` already exists and is non-empty, an error is thrown when writing to `file/html`, at least in min.io.  I'm not sure how this was ever allowed, it makes sense for it to be an error.  I'm not 100% sure that it is also an error in AWS but it really should be.

Note that if `file` is *empty* (0-bytes) then creating `file/html` *is* allowed.  This seems to be one of the accommodations that was made to emulate empty directories.

ericphanson: ah ok. That makes sense. I see in the next few lines we do
```julia
            is_aws(config) && k == "file" && continue
```
so I guess we are skipping that test on AWS, which makes sense since it sounds like it should fail-- looks like I introduced that check in #163. Before that, it was https://github.com/JuliaCloud/AWSS3.jl/blame/8ace53a78109d040d8686d5c67f296396f1fec11/test/awss3.jl#L142-L151 -- not sure why that workd? But it was in the tests for years and I don't think they were failing.

ExpandingMan: Yeah, I really don't know, the situation is very confusing across multiple versions of both minio and AWS, and I'm rather doubtful all the changes made to AWS are even documented.  Regardless: it was definitely failing in some cases, it doesn't seem as if it should succeed, doesn't seem appropriate to test it.

PR #199, opened 2021-08-17
Disallow empty version string

### LINKED ISSUE #181: Drop support for using empty string to get the latest S3 object version
As of https://github.com/JuliaCloud/AWSS3.jl/pull/161 we should be using `nothing` when trying to retrieve the latest version of an S3 object. In the past we supported using empty string which should be deprecated.

christopher-dG:
Closes #181, I think.

Alternatively, we could make this a deprecation since I don't think we did that in 0.8.x.

### DIFF
--- a/test/awss3.jl
+++ b/test/awss3.jl
@@ -223,6 +223,39 @@ function awss3_tests(config)
         @test !s3_exists(bucket_name, "testdir")
     end
 
+    @testset "Version is empty" begin
+        # Create the file to ensure we're only testing `version`
+        k = "version_empty.txt"
+        s3_put(config, bucket_name, k, "v1")
+        s3_put(config, bucket_name, k, "v2")
+
+        if is_aws(config)
+            @test_throws AWSException s3_get(bucket_name, k; version="")
+            @test_throws AWSException s3_get_meta(bucket_name, k; version="")
+            @test_throws AWSException s3_exists(bucket_name, k; version="")
+            @test_throws AWSException s3_delete(bucket_name, k; version="")
+        else
+            # Using an empty string as the version returns the latest version
+            @test s3_get(bucket_name, k; version="") == "v2"
+            @test s3_get_meta(bucket_name, k; version="") isa AbstractDict
+            @test s3_exists(bucket_name, k; version="")
+            @test s3_delete(bucket_name, k; version="") == UInt8[]
+        end
+    end
+
+    @testset "Version is nothing" begin
+        # Create the file to ensure we're only testing `version`
+        k = "version_nothing.txt"
+        s3_put(config, bucket_name, k, "v1")
+        s3_put(config, bucket_name, k, "v2")
+
+        # Using an empty string as the version returns the latest version
+        @test s3_get(bucket_name, k; version=nothing) == "v2"
+        @test s3_get_meta(bucket_name, k; version=nothing) isa AbstractDict
+        @test s3_exists(bucket_name, k; version=nothing)
+        @test s3_delete(bucket_name, k; version=nothing) == UInt8[]
+    end
+
     if is_aws(config)
         @testset "Empty and Delete Bucket" begin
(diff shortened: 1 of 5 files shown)


christopher-dG:
bors try

christopher-dG:
bors try

mattBrzezinski (approved):

test/awss3.jl
```diff
@@ -205,4 +205,13 @@ function awss3_tests(config)
             @test_throws AWS.AWSException s3_delete_bucket(config, bucket_name)
         end
     end
+
+    @testset "Empty version string" begin
```
mattBrzezinski, line 209: Could you add in a test for `S3Path` w/ an empty version here too?

omus:
> Alternatively, we could make this a deprecation since I don't think we did that in 0.8.x.

The nice thing to do IMO is we deprecate this nicely in a 0.8.9 backport and drop support for `""` in 0.9

test/s3path.jl
```diff
@@ -535,8 +537,7 @@ function s3path_tests(config)
             str_v1 = string(versioned_path_v1)
             roundtripped_v1 = S3Path(str_v1; config=config)
             @test isequal(versioned_path_v1, roundtripped_v1)
-            @test str_v1 ==
-                  "s3://" * bucket_name * "/" * key_version_file * "?versionId=" * v1
+            @test str_v1 == "s3://$(bucket_name)/$(key_version_file)?versionId=$(v1)"
```
omus, line 540: May want a test where we do `"s3://...?versionId="`

omus:
@christopher-dG is on vacation so I'll try to get this PR moving. I'll be trying to proceed with dropping support for treating `""` as `nothing` here and have a follow up PR which deprecates the behaviour for a 0.8 backport.

omus:
bors try

omus:
I was trying to determine the syntax of the "versionId" string and couldn't find any documentation. So I pulled down all the versions from a bucket and here were the results:
```julia
julia> lines = readlines("versions.txt")  # Extracted "VersionId" fields from `aws s3api list-object-versions`

julia> length(lines)
599220

julia> mapreduce(length, union, lines)
1-element Vector{Int64}:
 32

julia> String(sort!(reduce(union, lines)))
".0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz"
```
Based upon these results we could validate a proper "versionId" with the regex `r"^[0-9a-zA-Z\._]{32}$"`

omus:
bors try

omus:
Okay, so I discovered something that may change our approach here. Passing in `"versionId" = ""` is supported by the underlying AWS S3 API calls and returns the latest version of the file. The reason this is important is that I don't think we should go out of our way to check for `""` and fail when we find it. Instead, we should recommend using `nothing` and if someone uses `""` it can just rely on the API behaviour.

If however we do validate the versionId using the regex mentioned above then I think we would just disable using `""`

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
Ready for review. I made a backport PR of this here: https://github.com/JuliaCloud/AWSS3.jl/pull/202

omus:
bors try

omus:
bors try

omus:
Failure looks unrelated to me.

omus:
Rebased to resolve conflicts

omus:
bors try

mattBrzezinski (approved):

mattBrzezinski:
We really need to find a better solution to CI testing here. This is quite annoying, lets try again!

bors r+

christopher-dG (commented):
Sorry for late review, but I've just looked over your changes and all is good 👍

PR #200, opened 2021-08-21
Refactor some version related code to reduce redundancy

omus:
- Avoid separate calls to `S3` functions in favour of building up arguments
- Use `params` instead of `args` as this matches the `AWS.S3` function definitions and `args` should be reserved for use by the function for slurping
- Reduce redundant code in `S3Path` printing. Should reduce the possibility of inconsistencies being introduced

Split out from #199

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -105,9 +105,9 @@ function s3_get(
     kwargs...,
 )
     @repeat 4 try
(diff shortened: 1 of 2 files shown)


omus:
bors try

omus:
bors try

omus:
I'm proceeding with merging this as it should be non-controversial

PR #201, opened 2021-08-21
Format code to be readable and work with JuliaFormatter

omus:
Follow up to https://github.com/JuliaCloud/AWSS3.jl/pull/190 which I found butchered some of the code here. I mostly tried to revise some code where possible to work nicely with the formatter.

In a few cases where I believed the formatter to be wrong I filed issues and disabled it (seems worse to live with code we think is less readable).

Finally, the tests required some additional changes including renaming variables which I think resulted in much nicer tests.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -125,8 +125,7 @@ function s3_get(
 
         return S3.get_object(bucket, path, params; aws_config=aws, kwargs...)
     catch e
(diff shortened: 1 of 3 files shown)


omus:
bors try

mattBrzezinski (approved):
LGTM!

omus:
bors try

mattBrzezinski (approved):

Issue #170, opened 2021-07-14
Forward keyword arguments for s3_list_objects


tpgillam:
[This](https://github.com/JuliaCloud/AWSS3.jl/blob/master/src/AWSS3.jl#L458) dispatch for `s3_list_objects` is defined, which gives a default value for the aws config:
```
s3_list_objects(a...) = s3_list_objects(global_aws_config(), a...)
```
However, this only forwards positional arguments, not keyword arguments. This means that if one wishes to specify e.g. `delimeter` [here](https://github.com/JuliaCloud/AWSS3.jl/blob/master/src/AWSS3.jl#L407):
```
function s3_list_objects(aws::AbstractAWSConfig, bucket, path_prefix=""; delimiter="/", max_items=nothing, kwargs...)
```
then one cannot use the default value for `aws`.

I think the fix would be to add `; kwargs...` on both sides of the first dispatch?

PR #211, opened 2021-09-16
impelement `diskusage`

ExpandingMan:
I put functionality for checking the total "directory" size in `stat`, though notably this is *NOT* consistent with what happens on actual filesystems, in which `stat` (and pretty much everything else) only return the block size, not the actual directory size.  The corresponding linux command is `du -s`, but we don't seem to have an equivalent to this in `FilePathsBase`.  This conversation might therefore be better continued over there in which perhaps we should discuss whether we need a new function for getting the file size.

Since in pretty much all cases the actual size is what you really want, I figured it was ok to just stick it in there, though another concern is that this operation can be slow on very large directories.


ericphanson:
This seems like useful functionality but I’m a little bit concerned about `stat` in generic code doing something confusing (ie different from local paths) or surprisingly slow. So I wonder if it should just have a different name.

Is there a case where the person writing the code doesn’t know if a given path is meant to be a directory or not, and would like recursive stat if it is, and file stat otherwise? It’s hard for me to see it, because even something like “copy if it’s small, error otherwise” might need a check for whether or not to do the copy recursively.

I think it also could be confusing because exactly what is or isn’t a directory isn’t totally unambiguous (prefixes don’t actually have to be delimited by `/`, though we stick to that for S3Paths) so having different behaviour depending on whether or not AWSS3 thinks it’s a directory is a bit iffy in my opinion. But at least for empty directories the answer is the same I think, 0 bytes, though I’m not sure about the timestamp. So maybe it’s not so ambiguous.

ExpandingMan:
Yeah, I mostly agree with that.  My main concern is that this just isn't what `stat.size` means for directories, as much as I might like it to be.  I also think that if I take the size out of `stat` I also need to take the `last_modified` out because it seems crazy to iterate over a list for that (to me at least).

Ok, what should we call it then?  There's no function for this in `FilePathsBase` is there?  If not it should probably be added.  Of course, the simplest solution would be to just call it its own thing, though I don't entirely know what to call it since it would be really nice if it works on `S3Path` and I don't think `S3Path` has any exposed functions outside of the `AbstractPath` interface.

ExpandingMan:
For the time being at least, I have exported `s3_directory_stat` and added two new unexported functions `dirsize` and `dirlastmodified`.  Ideally, I think something like these should be added to the `AbstractPath` interface, but in the meantime I'd like to get this PR in a state where it can be merged and the details can be worried about later.

src/AWSS3.jl
```diff
@@ -603,8 +604,30 @@ function s3_list_objects(
...
 
-s3_list_objects(a...) = s3_list_objects(global_aws_config(), a...)
+"""
+    s3_directory_stat([::AbstractAWSConfig], bucket, path)
+
+Determine the properties of an S3 "directory", size and time of last modification, that cannot be determined
+directly with the standard AWS API.  This returns a tuple `(s, tmlast)` where `s` is the size in bytes, and
+`tmlast` is the time of the latest modification to a file within that directory.
+"""
+function s3_directory_stat(aws::AbstractAWSConfig, bucket, path)
```
mattBrzezinski, line 616: Can we type these params?

src/AWSS3.jl
```diff
@@ -603,8 +604,30 @@ function s3_list_objects(
...
+    s = 0
```
mattBrzezinski, line 617: And give some more descriptive names?

src/AWSS3.jl
ExpandingMan: That is generally my preference, but I was following the conventions already existing in this package.

src/AWSS3.jl
ExpandingMan: This is actually following some of the already-existing code above it, but I have added a comment.

src/AWSS3.jl
ExpandingMan: I added them for this anyway.  May be worth going back one day and cleaning this up in the rest of the package.

ExpandingMan:
Btw, the thing where the style linter constantly changes its mind about how you should indent function arguments based on exactly how many characters there are is infuriating.  The other stuff I at least ought to be able to spot myself and watch for as I type, but to follow that fiddly indent thing I'd basically have to have my editor settings conform to it.  I can't imagine that even people who follow this style everywhere find this very easy to deal with unless they have set their editor to exactly the right value.

Is there an issue for the linter that can be opened somewhere?  It's hard for me to imagine anyone being happy with the way that's currently working.

mattBrzezinski:
> Btw, the thing where the style linter constantly changes its mind about how you should indent function arguments based on exactly how many characters there are is infuriating. The other stuff I at least ought to be able to spot myself and watch for as I type, but to follow that fiddly indent thing I'd basically have to have my editor settings conform to it. I can't imagine that even people who follow this style everywhere find this very easy to deal with unless they have set their editor to exactly the right value.
> 
> Is there an issue for the linter that can be opened somewhere? It's hard for me to imagine anyone being happy with the way that's currently working.

https://github.com/domluna/JuliaFormatter.jl/

Which follows,

https://github.com/invenia/BlueStyle (at least this reviewdog does)

ExpandingMan:
I've been thinking about @rofinn 's response here.

Another possibility here is that we don't attempt to do anything like `dirsize`, but instead implement a version of `walkpath` that returns `Stat` structs rather than paths.  This would allow users to perform operations on these (such as summing them) without generating a huge number of unnecessary HTTP calls.  In that case we can just admit that AWSS3 needs that extra function which is not in the `AbstractPath` interface and make it clear in the docs.

As this is more general, I'm leaning toward this solution instead of what I've done here.  Thoughts?

ExpandingMan:
Ok, we are just waiting on a tag [here](https://github.com/rofinn/FilePathsBase.jl/pull/137) and then this will be an official part of the `AbstractPath` API.

src/AWSS3.jl
```diff
@@ -603,8 +604,32 @@ function s3_list_objects(
         end
     end
 end
+s3_list_objects(a...; kw...) = s3_list_objects(global_aws_config(), a...; kw...)
```
ericphanson, line 607: fixes https://github.com/JuliaCloud/AWSS3.jl/issues/170!

ExpandingMan: :slightly_smiling_face:

ExpandingMan:
Ok, as far as I know this should be all done now.

ericphanson (commented):
looks good to me, but probably we should reexport `diskusage, and export `lastmodified`?

also, I think we could use some tests for `lastmodified` (both cases), and the file case for `diskusage`.

ericphanson:
bors try

ExpandingMan:
Perhaps awkwardly, at the moment this package *ONLY* exports `S3Path` plus the `s3_` specific functions.  Should we add a bunch of exports for `FilePathsBase` stuff?  Alternatively we could just ask users to use `FilePathsBase` as most of its functionality extends `Base`.

`lastmodified` seems pretty hard to test, but I'll try to cook something up...

ExpandingMan:
I added a very basic test for `lastmodified` including one for a directory.  It's not a great test because I didn't want to risk it failing under weird circumstances (I even considered checking if they were on the same day, but I don't want it to fail at `00:00`!).  At least now the code path is exercised.

I've given exports some more thought, in my opinion there are 2 good options here:
1. Do nothing.
2. Reexport all of `FilePathsBase`.

The thinking is that we currently only export `S3Path` and `s3_` functions, exporting only `diskusage` and `lastmodified` would be bizarre.  I tend to think that this package, while clearly very important if you need it, is a bit too application specific to export generic-looking names.  When I use this package I almost always use `S3Path` *only*, and I think this is a good approach because it keeps a lot of S3-specific code out of packages that use it.

Anyway, please think about what you'd prefer for exports and I'll put it in, then this should be done.

test/s3path.jl
```diff
@@ -156,6 +156,19 @@ function test_s3_properties(ps::PathSet)
...
+        write(fp3 / "testfile1.txt", strs[1])
+        write(fp3 / "testfile2.txt", strs[2])
+        write(fp3 / "inner" / "testfile3.txt", strs[3])
+        @test AWSS3.diskusage(fp3) == sum(ncodeunits.(strs))
+
+        # we deliberately pick an older file to compare to so we
+        # can be confident timestamps are different
+        @test AWSS3.lastmodified(fp3) > AWSS3.lastmodified(ps.foo)
+
+        rm(S3Path(ps.root.bucket, "/another/"); recursive=true)  # otherwise subsequent tests may fail
```
ericphanson, line 171: should use try-finally to make sure we hit the `rm`:
```suggestion
        try
            fp3 = S3Path(ps.root.bucket, "/another/testdir/")
            strs = ["what up", "what else up", "what up again"]
            write(fp3 / "testfile1.txt", strs[1])
            write(fp3 / "testfile2.txt", strs[2])
            write(fp3 / "inner" / "testfile3.txt", strs[3])
            @test AWSS3.diskusage(fp3) == sum(ncodeunits.(strs))
    
            # we deliberately pick an older file to compare to so we
            # can be confident timestamps are different
            @test AWSS3.lastmodified(fp3) > AWSS3.lastmodified(ps.foo)
        finally
            rm(S3Path(ps.root.bucket, "/another/"); recursive=true)  # otherwise subsequent tests may fail
        end
```

ericphanson:
Ok, I agree we shouldn't selectively re-export FilesPathBase. I think then if we do want to re-export FilesPathBase that should be a separate PR-- no need to let this get bogged down with that decision.

bors try

ericphanson:
can you bump the patch version? Then I can tag right after merging

ericphanson:
bors merge

Issue #219, opened 2021-09-30
fix deprecations from new `AWS.Response` type


ExpandingMan:
This is just a reminder that we need to fix some deprecations due to the recent addition of the `AWS.Response` type.

In particular, we need to use
```julia
@service s3 use_response_type=true
```
because a deprecated `return_raw` option is used somewhere.

nickrobinson251:
just came to open this issue, as it broke some test in private code (tbf the test checking "no warnings" in too general a way), but also is just clogging up test logs, because now there are deprecation warnings:
```julia
julia> using AWSS3

julia> s3p = S3Path(...");

julia> s3_get_file(s3p.bucket, s3p.key, "local_foo.txt")
┌ Warning: The keyword `return_stream` is deprecated, use `@service S3 use_response_type=true` instead to return an `AWS.Response` allowing for streaming access via `response.io`.
│   caller = ip:0x0
└ @ Core :-1
```

PR #220, opened 2021-10-05
fixed deprecations from new AWS Response type

ExpandingMan:
Fixes deprecations (#219) from recent addition of AWS response type.  Since AWS.jl went to such great lengths not to break with that update (a mistake in my opinion, but I assume they are gearing up for a breaking change), this PR is non-breaking.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -52,7 +52,7 @@ using UUIDs
 using URIs
 using Compat: @something
 
-@service S3
+@service S3 use_response_type = true
 
 const SSDict = Dict{String,String}
 const AbstractS3Version = Union{AbstractString,Nothing}
@@ -80,8 +80,6 @@ from `path` in `bucket`.
 - `version=`: version of object to get.
 - `retry=true`: try again on "NoSuchBucket", "NoSuchKey"
                 (common if object was recently created).
-- `raw=false`:  return response as `Vector{UInt8}`
-                (by default return type depends on `Content-Type` header).
 - `byte_range=nothing`:  given an iterator of `(start_byte, end_byte)` gets only
     the range of bytes of the object from `start_byte` to `end_byte`.  For example,
     `byte_range=1:4` gets bytes 1 to 4 inclusive.  Arguments should use the Julia convention
@@ -101,14 +99,14 @@ function s3_get(
     path;
     version::AbstractS3Version=nothing,
     retry::Bool=true,
-    byte_range::Union{Nothing,AbstractVector}=nothing,
     raw::Bool=false,
+    byte_range::Union{Nothing,AbstractVector}=nothing,
     headers::AbstractDict{<:AbstractString,<:Any}=Dict{String,Any}(),
     return_stream::Bool=false,
     kwargs...,
 )
     @repeat 4 try
-        params = Dict{String,Any}("return_raw" => raw, "return_stream" => return_stream)
+        params = Dict{String,Any}()
         if version !== nothing
             params["versionId"] = version
         end
@@ -124,7 +122,14 @@ function s3_get(
             params["headers"] = headers
         end
 
-        return S3.get_object(bucket, path, params; aws_config=aws, kwargs...)
+        r = S3.get_object(bucket, path, params; aws_config=aws, kwargs...)
+        return if raw
+            r.body
+        elseif return_stream
+            r.io
+        else
+            parse(r)
+        end
     catch e
         #! format: off
         # https://github.com/domluna/JuliaFormatter.jl/issues/459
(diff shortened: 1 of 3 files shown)


src/AWSS3.jl
```diff
@@ -80,8 +80,6 @@ from `path` in `bucket`.
 - `version=`: version of object to get.
 - `retry=true`: try again on "NoSuchBucket", "NoSuchKey"
                 (common if object was recently created).
-- `raw=false`:  return response as `Vector{UInt8}`
```
morris25, line 83: Why are we removing this but keeping the kwarg?

src/AWSS3.jl
```diff
@@ -124,7 +122,14 @@ function s3_get(
             params["headers"] = headers
         end
 
-        return S3.get_object(bucket, path, params; aws_config=aws, kwargs...)
+        r = S3.get_object(bucket, path, params; aws_config=aws, kwargs...)
+        return if raw
```
morris25, line 126: It looks like `return_stream` currently takes precedence over `raw`.

```julia
julia> s3_get(global_aws_config(), bucket, file, raw=true)
52405-element Vector{UInt8}:
...

julia> s3_get(global_aws_config(), bucket, file, return_stream=true)
BufferStream() bytes waiting:52405, isopen:false

julia> s3_get(global_aws_config(), bucket, file, return_stream=true, raw=true)
BufferStream() bytes waiting:52405, isopen:false
```

src/AWSS3.jl
```diff
@@ -124,7 +123,14 @@ function s3_get(
             params["headers"] = headers
         end
 
-        return S3.get_object(bucket, path, params; aws_config=aws, kwargs...)
+        r = S3.get_object(bucket, path, params; aws_config=aws, kwargs...)
+        return if return_stream
+            r.io
```
omus, line 128: Will probably break things as this now returns a `IOBuffer` instead of the `BufferStream`. What we could do is pass in `"response_stream" => Base.BufferStream()` and then close it before returning it.

ExpandingMan:
I just want to be completely clear on what you are suggesting here, you want this to create a breaking release?  Obviously most of what I did was done to keep this from breaking.  This will of course need a breaking change at some point, but I wasn't sure whether everyone would go along with it before whenever AWS.jl 2.0 happens.

I'm going to commit your suggestions... so everyone is satisfied that this is breaking now right?

src/AWSS3.jl
ExpandingMan: Where does that argument get passed to?

ExpandingMan:
Ok, I was initially rather confused by what you had done, I now realize that nothing you suggested was breaking.

I have committed or otherwise implemented all of your suggestions but one, see comment above.

ExpandingMan:
I don't have the link right now but the AWS.jl devs are aware that there is going to have to be a breaking release of that soon.

Therfore, my preference would be to merge this PR, and then there is going to have to be a comprehensive overhaul once the new AWS.jl is tagged.

Everything is resolved but the `BufferStream` issue, are you sure that's breaking?  `BufferStream` breaks rather spectacularly.

src/AWSS3.jl
```diff
@@ -355,16 +364,16 @@ s3_copy(a...; b...) = s3_copy(global_aws_config(), a...; b...)
...
-                                <CreateBucketConfiguration
-                                            xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
-                                    <LocationConstraint>$(aws.region)</LocationConstraint>
-                                </CreateBucketConfiguration>
-                            """
+            <CreateBucketConfiguration
+            xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+            <LocationConstraint>$(aws.region)</LocationConstraint>
+            </CreateBucketConfiguration>
+            """
```
omus, line 376: Should be within then 92-character line limit
```suggestion
                <CreateBucketConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
                    <LocationConstraint>$(aws.region)</LocationConstraint>
                </CreateBucketConfiguration>
                """
```

src/AWSS3.jl
```diff
@@ -416,18 +426,20 @@ s3_put_cors(a...; b...) = s3_put_cors(AWS.global_aws_config(), a...; b...)
...
 function s3_enable_versioning(aws::AbstractAWSConfig, bucket, status="Enabled"; kwargs...)
     versioning_config = """
-        <VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
-            <Status>$status</Status>
-        </VersioningConfiguration>
-        """
+    <VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+    <Status>$status</Status>
+    </VersioningConfiguration>
+    """
```
omus, line 432:
```suggestion
        <VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
            <Status>$status</Status>
        </VersioningConfiguration>
        """
```

ExpandingMan:
Ok, should be all wrapped up now.

ExpandingMan:
If there are no objections (and assuming it passes bors), I'd prefer to just merge this as is if you are willing.

There will have to be a major overhaul when AWS.jl tags a breaking release, with lots of opportunities to clean things up for the better.

PR #229, opened 2022-01-21
add no-op `S3Path(::S3Path)` constructor

kleinschmidt:
This is handy when you want to make sure you're dealing with an S3Path when you may get a string or an S3Path.

Issue #227, opened 2021-12-02
CI test failure:  MethodError: no method matching parse(::Type{S3Path{Nothing}}, ::S3Path{Nothing})


morris25:
The following error occurs several times in [current CI pipelines](https://github.com/JuliaCloud/AWSS3.jl/runs/4390051615?check_suite_focus=true):
```
MethodError: no method matching parse(::Type{S3Path{Nothing}}, ::S3Path{Nothing})
  Closest candidates are:
    parse(::Type{P}, ::AbstractString; kwargs...) where P<:AbstractPath at ~/.julia/packages/FilePathsBase/qgXdE/src/path.jl:77
  Stacktrace:
    [1] _broadcast_getindex_evalf
      @ ./broadcast.jl:670 [inlined]
    [2] _broadcast_getindex
      @ ./broadcast.jl:653 [inlined]
    [3] getindex
      @ ./broadcast.jl:597 [inlined]
    [4] copy
      @ ./broadcast.jl:899 [inlined]
    [5] materialize
      @ ./broadcast.jl:860 [inlined]
    [6] readpath(p::S3Path{Nothing}; join::Bool, sort::Bool)
      @ FilePathsBase ~/.julia/packages/FilePathsBase/qgXdE/src/path.jl:713
    [7] readpath
      @ ~/.julia/packages/FilePathsBase/qgXdE/src/path.jl:708 [inlined]
    [8] macro expansion
      @ /opt/hostedtoolcache/julia/1.7.0/x64/share/julia/stdlib/v1.7/Test/src/Test.jl:445 [inlined]
    [9] macro expansion
      @ ~/work/AWSS3.jl/AWSS3.jl/test/s3path.jl:59 [inlined]
   [10] macro expansion
      @ /opt/hostedtoolcache/julia/1.7.0/x64/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
   [11] test_s3_readpath(p::PathSet{S3Path{Nothing}})
      @ Main ~/work/AWSS3.jl/AWSS3.jl/test/s3path.jl:57
```
Likely causes are changes in FilePathsBase or the S3 bucket failing to be created/being deleted by another pipeline.  I haven't investigated the error fully but the last successful pipeline was 15 days ago. FilePathsBase was at v0.9.15 on the last successful run. It is now at v0.9.17.

rofinn:
[...]

Issue #226, opened 2021-11-16
Should S3Path stop using the FilePathsBase interface?


ericphanson:
It provides a lot of methods for free, but an S3Path is not like a file system path in a lot of ways, and the abstractions break down in frustrating ways. I think it might be better for AWSS3 to define it's own methods for most things, so that things that are not explicitly supported by AWSS3 are not supported at all, rather than supported in a somewhat broken way (which you might not notice at first but can bite you later-- for example `cp`, which works until you have nested directories, ref #225).

inspired by https://github.com/rofinn/FilePathsBase.jl/issues/128#issuecomment-970177142

ericphanson:
I think it would make sense for an S3Path to roughly mimic the AWS console commands, `aws s3 cp ...`, `aws s3 ls ...`, etc. So, following https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3/index.html, I think these operations make definitely sense:

* `cp`
    * `write` and `read`, which are like `aws s3 cp - s3://...` and `aws s3 cp s3://... -`
* `mv`
* `rm`
* `readdir` (analogous to `ls`)
* `sync`

(I don't think we need `rb` or `mb`, at least currently).

I think they should match Base's semantics though, not the AWS console ones, e.g. `cp` is recursive by default.

And then some level of path-manipulation operations could make sense too (manipulating S3Path objects, i.e. *not* as an API to interact with s3):

* `joinpath`
* `splitext`
* `relpath`
* `splitdir`

But they need to be pretty careful to follow s3's semantics for prefixes.

And I think these functions *don't* really make sense:

* `mkpath`, `mkdir`: maybe could define them as a no-op, ref #224
* `cmod`, `chown`, etc: don't want to be messing with permissions

And these are the tricky ones:

* `isfile`
* `isdir`
* `ispath`

which I don't know how to deal with.

hannahilea:
...our team has lost a _lot_ of time collectively to [diagnosing/working around/bug fixing/knowledge sharing] issues that boil down to 

> the abstractions break down in frustrating ways 

and
> supported in a somewhat broken way (which you might not notice at first but can bite you later [...)]

so I'm strongly in favor of this proposal.

rofinn:
[...]

PR #230, opened 2022-01-25
readdir must return strings for FilePathsBase compatibility

ExpandingMan:
Currently doing `readdir(p, join=true)` returns an array of paths.  This is a problem because FilePathsBase expects it to be strings in all cases.

This is breaking things pretty badly (e.g. even `rm` on directories is currently broken), so I'm hoping we can get this merged relatively quickly.

### DIFF
--- a/src/s3path.jl
+++ b/src/s3path.jl
@@ -582,7 +582,7 @@ function Base.readdir(fp::S3Path; join=false, sort=true)
         sort && sort!(results)
 
         # Return results, possibly joined with the root path if join=true
-        return join ? joinpath.(fp, results) : results
+        return join ? string.(joinpath.(fp, results)) : results
     else
         throw(ArgumentError("\"$fp\" is not a directory"))
     end


ExpandingMan:
This is a little worrying because we are also testing against what is apparently the wrong behavior... I don't *think* FilePathsBase has changed, but I'm not entirely sure.

bors try

ericphanson:
> I don't think FilePathsBase has changed, but I'm not entirely sure

Yep, FilePathsBase made a patch release that added `join=true` and also broke AWSS3: https://github.com/JuliaCloud/AWSS3.jl/issues/226#issuecomment-1004210528 (see also https://github.com/JuliaCloud/AWSS3.jl/issues/227).

We can make this change here*, but it would definitely need to be a breaking change for AWSS3.jl.

*: I think these are the wrong semantics, because it means `read(readdir(...;join=true)[1])` now throws, but it would at least be consistent with FilePathsBase. Not really sure if it's worth it or not. As expressed in #226, I think we should stop trying to fulfill the FilePathsBase interface altogether, which would also definitely be a breaking change.

ericphanson (changes requested):
Needs breaking version bump if we are going to do this

ExpandingMan:
In my opinion breaking with FilePathsBase would drastically diminish the usefulness of this package.

The current state of the package is rather broken, so this needs to be done until and unless you won't to re-implement everything.

bors try

ExpandingMan:
I'm not sure what's happening with bors.  I thought it got stalled out because I made another commit, but now it says it's already running and doesn't want to run.

ericphanson:
>I'm not sure what's happening with bors. I thought it got stalled out because I made another commit, but now it says it's already running and doesn't want to run.

I think Bors might be broken; last time I tried it in https://github.com/JuliaCloud/AWSS3.jl/pull/229 it timed out. Not sure what's going on though, maybe @mattBrzezinski can look into it?

> The current state of the package is rather broken, so this needs to be done until and unless you won't to re-implement everything.

Well, we could just restrict compat to non-broken FilePathsBase? This PR could do that instead. Ref https://github.com/JuliaCloud/AWSS3.jl/issues/226#issuecomment-1005088875

ExpandingMan:
Restricting the FilePathsBase versoin without fixing this seems like a terrible idea, because then this package will just get stuck on an effectively unmaintained version of it.  If AWSS3 drops FilePathsBase, then so be it, but before that has happened restricting it seems like a terrible idea.

I also don't think adding a method to `Base.parse` makes much sense, the FilePathsBase behavior is logical: its return type does not depend on `join`.

So yes, even though I disagree that it is desirable, there is a case to be made for dropping FilePathsBase, but in the meantime, it seems better to try to keep things working properly.

ericphanson:
> So yes, even though I disagree that it is desirable, there is a case to be made for dropping FilePathsBase, but in the meantime, it seems better to try to keep things working properly.

Ok, but breaking the fact that you can `read` the outputs `readdir(...;join=true)` is not keeping things working properly! That's like a pretty key bit of using S3Paths in generic code. So if we really want to do that for the sake of consistency, this needs to be a breaking release. Or we can add the no-op `parse` method to workaround the breakage from https://github.com/rofinn/FilePathsBase.jl/pull/152/files#diff-4e2b33a7304f53abf36fe9499936056b993077836e4b72a5c5ae016d8c6d6d1fR713.

ExpandingMan:
I'm re-thinking my stance against extending `Base.parse`.  I still think it is ultimately a bad idea, but I think we are headed to some kind of reckoning, so maybe it makes sense to merely patch things up before that happens.

Btw, please let me know what you think about [my draft PR to FilePathsBase](https://github.com/rofinn/FilePathsBase.jl/pull/159) in that thread.

test/s3path.jl
```diff
@@ -245,19 +245,21 @@ end
 function verify_files(path::S3Path)
     @test readdir(path) == ["emptydir/", "subdir1/", "test_01.txt"]
     @test readdir(path; join=true) ==
-        [path / "emptydir/", path / "subdir1/", path / "test_01.txt"]
+        string.([path / "emptydir/", path / "subdir1/", path / "test_01.txt"])
```
omus, line 248:
```suggestion
        String[path / "emptydir/", path / "subdir1/", path / "test_01.txt"]
```

test/s3path.jl
```diff
@@ -245,19 +245,21 @@ end
...
     @test readdir(path / "emptydir/") == []
     @test readdir(path / "emptydir/"; join=true) == []
     @test readdir(path / "subdir1/") == ["subdir2/", "test_02.txt", "test_03.txt"]
-    @test readdir(path / "subdir1/"; join=true) == [
+    @test readdir(path / "subdir1/"; join=true) ==
+        string.([
```
omus, line 253:
```suggestion
    @test readdir(path / "subdir1/"; join=true) == String[
```

test/s3path.jl
```diff
@@ -245,19 +245,21 @@ end
...
         path / "subdir1/" / "subdir2/",
         path / "subdir1/" / "test_02.txt",
         path / "subdir1/" / "test_03.txt",
-    ]
+    ])
```
omus, line 257:
```suggestion
    ]
```

test/s3path.jl
```diff
@@ -245,19 +245,21 @@ end
...
     @test readdir(path / "subdir1/subdir2/") == ["subdir3/", "test_04.txt"]
-    @test readdir(path / "subdir1/subdir2/"; join=true) == [
+    @test readdir(path / "subdir1/subdir2/"; join=true) ==
+        string.([
```
omus, line 260:
```suggestion
    @test readdir(path / "subdir1/subdir2/"; join=true) == String[
```

test/s3path.jl
```diff
@@ -245,19 +245,21 @@ end
...
         path / "subdir1/subdir2/" / "subdir3/", path / "subdir1/subdir2/" / "test_04.txt"
-    ]
+    ])
```
omus, line 262:
```suggestion
    ]
```

test/s3path.jl
```diff
@@ -269,11 +271,12 @@ function verify_files(path::AbstractPath)
     @test readdir(path / "emptydir/") == []
     VERSION >= v"1.4.0" && @test readdir(path / "emptydir/"; join=true) == []
     @test readdir(path / "subdir1/") == ["subdir2", "test_02.txt", "test_03.txt"]
-    VERSION >= v"1.4.0" && @test readdir(path / "subdir1/"; join=true) == [
+    VERSION >= v"1.4.0" && @test readdir(path / "subdir1/"; join=true) ==
+        string.([
```
omus, line 275:
```suggestion
    VERSION >= v"1.4.0" && @test readdir(path / "subdir1/"; join=true) == String[
```

test/s3path.jl
```diff
@@ -269,11 +271,12 @@ function verify_files(path::AbstractPath)
...
         path / "subdir1" / "subdir2",
         path / "subdir1" / "test_02.txt",
         path / "subdir1/" / "subdir1/test_03.txt",
-    ]
+    ])
```
omus, line 279:
```suggestion
    ]
```

omus (commented):
Test formatting changes

ExpandingMan:
Sorry, was busy.  Let's try running again

bors try

PR #232, opened 2022-02-03
Set FilePathsBase compatibility to 0.9.11 - 0.9.15

### LINKED ISSUE #227: CI test failure:  MethodError: no method matching parse(::Type{S3Path{Nothing}}, ::S3Path{Nothing})

omus:
Fixes #227 by avoiding using FilePathsBase versions that require `readdir` when using `join=true` to return strings. As changing the return type could be breaking for some users we'll just limit the versions of FilePathsBase supported here for now for this release. For AWSS3.jl version 0.10 we'll correct that problem in this package and start using a minimum version of FilePathsBase 0.9.16

At first glance appears to address #226 but it does not as we are just temporarily setting an upperbound.

Also, we cannot use Pkg version ranges using a hypen as that requires a minimum version of Julia 1.4


omus:
bors try

omus:
I've managed to trigger bors today with a `bors r+` but not with `bors try`. I'll do some experimentation here

omus:
bors p=1

omus:
bors try

ericphanson:
if we want to take this approach, we will also need to edit the registry and add this cap to all previous versions of AWSS3 since `readdir( ;join)` was introduced, otherwise the resolver will just give you e.g. AWSS3 v0.9.2 + FilePathsBase v0.9.16

omus:
> if we want to take this approach, we will also need to edit the registry and add this cap to all previous versions of AWSS3 since `readdir( ;join)` was introduced, otherwise the resolver will just give you e.g. AWSS3 v0.9.2 + FilePathsBase v0.9.16

I'm good for doing that as those versions are broken with FilePathsBase 0.9.16

omus:
bors try-

omus:
bors try

omus:
I seem to have set the upper bound to the wrong version

omus:
bors try

ericphanson (approved):

omus:
I'll merge this in the morning and get a few of the previously stuck PRs merged

omus:
Hmm, looks like there is more investigation to do here. I'll try a re-run to validate the error is reproducible

omus:
bors try

omus:
Interesting. Let's try merging again:

bors r+

omus:
2/3 jobs have finished but the macOS job seems to have failed. I'll wait a little longer but I'll manually merge this if we see another failure. This change is quite simple and the failure are unrelated.

PR #233, opened 2022-02-04
Debug macOS segfault on CI

omus:
Seeing if I can narrow down where the macOS segfault is occurring on CI instances. I can't reproduce this locally.


omus:
bors try

omus:
I'm rapidly running out of time today. Since Julia 1.3 works well on macOS I think we can just test that for now to get the other PRs merged.

omus:
bors try-

PR #234, opened 2022-02-04
Work around macOS segfaults/stalls on CI

omus:
Noticed in #233 that macOS on Julia 1.3 passed while Julia 1 (1.7.1) would either stall out or segfault on CI only. It seems like this package hasn't been getting enough maintenance lately and this should help unblock things.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,8 +26,10 @@ jobs:
         arch:
           - x64
         exclude:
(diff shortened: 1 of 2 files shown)


omus:
bors try

omus:
bors try

omus:
bors try-

omus:
bors try

omus:
I forgot bors doesn't squash. I'll do that first

PR #245, opened 2022-03-28
readavailable(::FileBuffer) missing in FilePathsBase < 0.9.18

schlurp:


Project.toml
```diff
@@ -24,7 +24,7 @@ AWS = "1.63.1"
 ArrowTypes = "1.2"
 Compat = "3.29.0"
 EzXML = "0.9, 1"
-FilePathsBase = "=0.9.11, =0.9.12, =0.9.13, =0.9.14, =0.9.15"
+FilePathsBase = "=0.9.11, =0.9.12, =0.9.13, =0.9.14, =0.9.15, =0.9.18"
```
ericphanson, line 27:
```suggestion
FilePathsBase = "0.9"
```

ericphanson (approved):
This is the workaround I suggested in https://github.com/JuliaCloud/AWSS3.jl/pull/230#issuecomment-1021702549 and I’m still in favor :). We should just relax the compat since we have the workaround here instead of adding 0.9.18 specifically.

Project.toml
schlurp: Ok, wasn't aware that there has also been another no-op parse been implemented. Just needed a quickfix on my side since the upgrade of CSV broke my production code... ;)

ericphanson: the issue is that FilePathsBase 0.9.16 started relying on `parse` in `readdir` (ref https://github.com/rofinn/FilePathsBase.jl/pull/152) in a way that conflicted with https://github.com/JuliaCloud/AWSS3.jl/pull/129, so to temporarily solve the issue, we compat bounded strictly to 0.9.11-0.9.15 (ref https://github.com/JuliaCloud/AWSS3.jl/pull/232). But if we add the `parse` method here, then we don't need that strict compat, so we can get new FilePathsBase releases that have fixes like the `readavailable` one.

ericphanson:
bors try

src/s3path.jl
```diff
@@ -151,6 +151,10 @@ function Base.tryparse(
     return S3Path(path, root, drive, isdirectory, version, config)
 end
 
+function Base.parse(::Type{P}, pth::P; kwargs...) where P<:S3Path
+    return pth
+end
+
```
mattBrzezinski, line 157:
```suggestion
function Base.parse(::Type{P}, p::P; kwargs...) where {P<:S3Path}
    return p
end

```

mattBrzezinski (approved):
bors try

ericphanson:
bors try

ericphanson:
I'll merge tomorrow if there are no objections

Project.toml
```diff
@@ -24,7 +24,7 @@ AWS = "1.63.1"
 ArrowTypes = "1.2"
 Compat = "3.29.0"
 EzXML = "0.9, 1"
-FilePathsBase = "=0.9.11, =0.9.12, =0.9.13, =0.9.14, =0.9.15"
+FilePathsBase = "0.9"
```
iamed2, line 27:
```suggestion
FilePathsBase = "0.9.11"
```
Might as well set a minimum rather than re-allowing older versions, especially if we're all the way up at 0.9.18. This is what it was at before the change to this version of compat.

ericphanson:
bors merge

PR #255, opened 2022-06-29
Drop HTTP since its not a necessary explicit dependency

quinnj:
I'm not sure if the FilePathsBase compats are correct, so please correct me if those won't work.


quinnj:
bors try

quinnj:
bors try

quinnj:
Good catch; bumped.

quinnj:
bors try

.github/workflows/CompatHelper.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: julia-actions/setup-julia@latest
         with:
-          version: "1.3"
+          version: "1"
```
iamed2, line 13:
```suggestion
          version: "1.6"
```

We don't want CompatHelper suggesting changes that aren't compatible with the minimum Julia version (AFAICT from various GitHub and Slack conversations, the resolution algorithm is currently tied to the running version of Julia)

ericphanson (approved):
LGTM!

Issue #252, opened 2022-06-01
stat / s3_get_meta rely on case-sensitivity of headers


chris-b1:
Using a custom config for s3 compatible [DigitalOcean spaces](https://docs.digitalocean.com/products/spaces/), I get the following dictionary for `s3_get_meta`

```
julia> s3_get_meta(cfg, "mybucket", "mykey")
Dict{SubString{String}, SubString{String}} with 10 entries:
  "etag"                      => "\"b6b7e9e077600dd1dcab6fd8b7551b2e\""
  "vary"                      => "Origin, Access-Control-Request-Headers, Access-Control-Request-Method"
  "strict-transport-security" => "max-age=15552000; includeSubDomains; preload"
  "x-amz-request-id"          => "tx0000000000000901984d3-006297bd4c-319bec8f-nyc3c"
  "x-rgw-object-type"         => "Normal"
  "last-modified"             => "Wed, 01 Jun 2022 19:04:08 GMT"
  "date"                      => "Wed, 01 Jun 2022 19:26:04 GMT"
  "accept-ranges"             => "bytes"
  "content-length"            => "757248"
```

The `Base.stat` overload fails on this because it expects uppercase "Last-Modified"

```
julia> p = S3Path("s3://mybucket/mykey"; config=cfg)

julia> stat(p)
ERROR: KeyError: key "Last-Modified" not found
Stacktrace:
 [1] getindex(h::Dict{SubString{String}, SubString{String}}, key::String)
   @ Base .\dict.jl:481
 [2] stat(fp::S3Path{Spaces.SpacesConfig})
   @ AWSS3 C:\Users\chris.bartak\.julia\packages\AWSS3\D1gQl\src\s3path.jl:360
 [3] top-level scope
   @ REPL[23]:1
```

kleinschmidt:
I'm seeing the same thing on AWS S3 (all keys are lowercase).

kleinschmidt:
(This despite the fact that the docs for this endpoint clearly suggest that the headers should be "Last-Modified" etc.: https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_ResponseSyntax)

ericphanson:
This is https://github.com/JuliaCloud/AWS.jl/issues/547 (which also hits https://github.com/JuliaCloud/AWSS3.jl/issues/246)

Issue #246, opened 2022-04-08
`KeyError: key "ETag" not found` in `s3_multipart_upload`


ericphanson:
This is code that has run probably a couple dozen (or hundred?) times since I last touched it, but just now I guess AWS decided to not give an ETag back? No idea what went wrong.
```julia
ERROR: LoadError: KeyError: key "ETag" not found
Stacktrace:
 [1] getindex(h::Dict{SubString{String}, SubString{String}}, key::String)
   @ Base ./dict.jl:481
 [2] s3_upload_part(aws::AWS.AWSConfig, upload::OrderedCollections.LittleDict{Union{String, Symbol}, Any, Vector{Union{String, Symbol}}, Vector{Any}}, part_number::Int64, part_data::Vector{UInt8}; args::Dict{String, Any}, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ AWSS3 ~/.julia/packages/AWSS3/D1gQl/src/AWSS3.jl:841
 [3] s3_upload_part(aws::AWS.AWSConfig, upload::OrderedCollections.LittleDict{Union{String, Symbol}, Any, Vector{Union{String, Symbol}}, Vector{Any}}, part_number::Int64, part_data::Vector{UInt8})
   @ AWSS3 ~/.julia/packages/AWSS3/D1gQl/src/AWSS3.jl:829
 [4] s3_multipart_upload(aws::AWS.AWSConfig, bucket::SubString{String}, path::String, io::IOStream, part_size_mb::Int64; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ AWSS3 ~/.julia/packages/AWSS3/D1gQl/src/AWSS3.jl:884
 [5] s3_multipart_upload (repeats 2 times)
   @ ~/.julia/packages/AWSS3/D1gQl/src/AWSS3.jl:872 [inlined]
```

I guess we should retry if we don't get the ETag? Not really sure what's going on. The docs say if it's successful, it should be a 200 response w/ an ETag. Maybe it wasn't a 200 but we didn't throw in AWS.jl or something?

omus:
[...]

PR #260, opened 2022-07-15
Robustly handle case variation in return value of s3_get_meta

haberdashPI:
The case of keys in the return value `s3_get_meta` can vary. Sometime, in the same julia session, it will be "Content-Length" and other times it will be "content-length". :rage: I'm assuming this is an issue with S3 itself. 

To handle this situation I've written a small helper: `get_robust_case`. There are probably other places where it is useful, but I'm just fixing it for calls to `stat` here (where I often run into this problem).

closes https://github.com/JuliaCloud/AWSS3.jl/issues/252
closes https://github.com/JuliaCloud/AWSS3.jl/issues/246

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -58,6 +58,12 @@ const SSDict = Dict{String,String}
 const AbstractS3Version = Union{AbstractString,Nothing}
 const AbstractS3PathConfig = Union{AbstractAWSConfig,Nothing}
 
+# Utility function to workaround https://github.com/JuliaCloud/AWS.jl/issues/547
+function get_robust_case(x, key)
+    haskey(x, key) && return x[key]
+    return x[lowercase(key)]
+end
+
 __init__() = FilePathsBase.register(S3Path)
 
 # Declare new `parse` function to avoid type piracy
@@ -837,7 +843,7 @@ function s3_upload_part(
         kwargs...,
     )
 
-    return Dict(response.headers)["ETag"]
+    return get_robust_case(Dict(response.headers), "ETag")
 end
 
 function s3_complete_multipart_upload(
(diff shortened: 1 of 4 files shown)


kleinschmidt:
#252

ericphanson:
I think these seems like a reasonable workaround. Can you add it to the `ETag` call in `s3_upload_part` as well?

bors try

src/AWSS3.jl
```diff
@@ -58,6 +58,12 @@ const SSDict = Dict{String,String}
 const AbstractS3Version = Union{AbstractString,Nothing}
 const AbstractS3PathConfig = Union{AbstractAWSConfig,Nothing}
 
+# Utility function to workaround https://github.com/JuliaCloud/AWS.jl/issues/547
+function get_robust_case(x, key)
+    haskey(x, key) && return x[key]
+    return x[lowercase(key)]
```
ericphanson, line 64: I think we should switch the order, so the error makes more sense if there isn't a key at all (regardless of case). I.e. the KeyError should probably have the standard casing instead of the lowercase casing:
```suggestion
    lkey = lowercase(key)
    haskey(x, lkey) && return x[lkey]
    return x[key]
```

ericphanson:
We should test with the DownloadsBackend to catch these things way earlier. For this PR though, maybe it's enough to add unit tests for the new function.

test/awss3.jl
```diff
@@ -3,6 +3,15 @@ function awss3_tests(config)
         "ocaws.jl.test." * lowercase(Dates.format(now(Dates.UTC), df))
     end
 
+    @testset "Robust key selection" begin
+        lower_dict = Dict("foo-bar" => 1)
+        upper_dict = Dict("Foo-Bar" => 1)
+        @test AWSS3.get_robust_case(lower_dict, "Foo-Bar") == 1
+        @test AWSS3.get_robust_case(upper_dict, "Foo-Bar") == 1
+        @test AWSS3.get_robust_case(Dict(), "Foo-Bar")
```
ericphanson, line 11:
```suggestion
```

ericphanson:
bors merge

Issue #254, opened 2022-06-22
RFC: FilePaths2.jl


ExpandingMan:
Hello all.  I have, over the past couple of months, been extremely determined to completely solve what I consider to be an unfortunate state of affairs with `S3Path` from AWSS3.jl.  You've probably all seen my speculative PR's having to do with adding traits to accommodate this.  I have recently largely rewritten AbstractTrees.jl partially in service of this goal.

I've ultimately given the situation a lot more thought and have come up with [FilePaths2.jl](https://gitlab.com/ExpandingMan/FilePaths2.jl).

The package is currently deliberately incomplete as I would like to solicit opinions.  FilePathsBase.jl could be overhauled based on my approach in FilePaths2.jl.  I'd be glad to do that, but obviously I do not know what the maintainers would think of that.  My current repo of FilePaths2.jl is meant as an example of what I intend to do, please see the README for a more careful exposition of what I did and why.

Would like to hear opinions about whether the maintainers would be receptive to moving AWSS3.jl to be based on this approach, whether that would be the result of changing FilePathsBase or a new package.  (Cross posting [here](https://github.com/rofinn/FilePathsBase.jl/issues/164))

PR #265, opened 2022-08-02
Use list_objects_v2 to check isdir

fchorney:
isdir for an s3 directory currently calls `s3_list_buckets` to check if a top level bucket is a directory. I ran into an issue where the role I was using didn't have permissions to call `list_buckets`.

This change essentially calls `S3.list_objects_v2` with `max-keys` set to `0` as we don't actually need to retrieve any keys. If the bucket exists and you have permissions for it, it will return with no error, but if the bucket doesn't exist we return false or raise an exception if some other error occurred. 

Not too sure if this is exactly how we want to go about this (re: the exceptions), so if you have any suggestions let me know. I don't think there should be any issues for using `S3.list_objects_v2` instead of `s3_list_buckets`.


rofinn:
Since this is just changing the behaviour for root bucket paths I think it's fine. You should add tests for each of the conditions though:

1. Normal `isdir` (probably already exists)
2. No such bucket error
3. Any other error

mattBrzezinski:
I think these changes are fine, but don't really solve the root problem here, which is to add the permissions needed to the role? We're now requiring users to have permissions to `ListObjects` on a bucket, instead of `ListBuckets`.

However, I do think those are the more correct permissions to require. I would also tag this as a `0.10.0` release instead of `0.9.X` as it could be breaking for some users.

fchorney:
> I think these changes are fine, but don't really solve the root problem here, which is to add the permissions needed to the role? We're now requiring users to have permissions to `ListObjects` on a bucket, instead of `ListBuckets`.
> 
> However, I do think those are the more correct permissions to require. I would also tag this as a `0.10.0` release instead of `0.9.X` as it could be breaking for some users.

You already need `ListObjects` to check `exists(fp)` which happens on a non root bucket, so in this case you would theoretically already have `ListObjects` if you're calling `readdir` or `isdir`

mattBrzezinski:
> You already need `ListObjects` to check `exists(fp)` which happens on a non root bucket, so in this case you would theoretically already have `ListObjects` if you're calling `readdir` or `isdir`

Oh rad, I missed that! Just needs test cases and it should be good to go!

Side thought, made we should include in our documentation the permissions users would need to run these functions?

test/runtests.jl
```diff
@@ -10,6 +10,9 @@ using FilePathsBase: /, join
 using FilePathsBase.TestPaths
 using UUIDs: uuid4
 using JSON3
+using Mocking
```
mattBrzezinski, line 13: Would be nice to order these alphabetically

test/s3path.jl
```diff
@@ -421,6 +421,31 @@ function s3path_tests(config)
         end
     end
 
+    @testset "isdir" begin
+        # Top level
+        path = S3Path("s3://$(bucket_name)"; config=config)
+        @test isdir(path) == true
+
+        # No Such Bucket
+        path = S3Path("s3://some_non_existent_bucket_7051649213"; config=config)
```
mattBrzezinski, line 430: We should make this an actual AWSException w/ the `NoSuchBucket` error so we're not dependent on someone not making this an actual bucket.

test/runtests.jl
fchorney: yeah I was thinking of doing that but it felt kind of out not related to this particular MR, but if we're fine with it I'll reorder all the usings.

fchorney:
bors try

mjram0s (approved):

test/s3path.jl
```diff
@@ -421,6 +421,42 @@ function s3path_tests(config)
...
+            nothing,
+        )
+        patch = @patch AWSS3.S3.list_objects_v2(args...; kwargs...) = throw(test_exception)
+
+        path = S3Path("s3://$(bucket_name)"; config=config)
+        apply(patch) do
+            @test_throws AWSException isdir(path)
+        end
+    end
+
```
mattBrzezinski, line 459:
```suggestion
    @testset "isdir" begin
        function _generate_exception(code)
            return AWSException(
                code,
                "",
                nothing,
                AWS.HTTP.Exceptions.StatusError(404, "", "", ""),
                nothing,
            )
        end

        const s3_path = S3Path("s3://$(bucket_name"; config=config)

        @testset "top level" begin
            @test isdir(s3_path) == true
        end
        
        @testset "no such bucket" begin
            test_exception = _generate_exception("NoSuchBucket")
            patch = @patch AWSS3.S3.list_objects_v2(args...; kwargs...) = throw(test_exception)

            apply(patch) do
                @test isdir(s3_path) == false
            end
        end

        @testset "other error thrown" begin
            test_exception = _generate_exception("TestException")
            patch = @patch AWSS3.S3.list_objects_v2(args...; kwargs...) = throw(test_exception)
            apply(patch) do
                @test_throws AWSException isdir(s3_path)
            end
        end
    end

```

fchorney: Modified it a bit since all the tests are for top level bucket access.

rofinn (approved):
Seems fine to me. I do think we should bump the minor release though.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "AWSS3"
 uuid = "1c724243-ef5b-51ab-93f4-b0a88ac62a95"
-version = "0.9.8"
+version = "0.9.9"
```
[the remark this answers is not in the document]
fchorney: I'm somewhat unsure if that's the case, what behaviour could someone depend on? Theoretically you already *need* `list_objects` permissions to do an `isdir` on a non root bucket. I guess this could potentially be an issue for someone that is only doing an `isdir` on a root bucket somehow? I'm fine to bump the minor if you still think that should be the case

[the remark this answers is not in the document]
fchorney: Bumped the version regardless

fchorney:
This is frustrating. `AWSS3.jl` depends on `Minio.jl` for tests, but `Minio.jl` depends on `AWSS3.jl` and its deps don't allow for anything above `0.9`

ExpandingMan:
I would have to not only push the change to `master` but actually *tag* the commit for this to resolve the issue here.  Can I get some assurances that this is what will be tagged as 0.10 before I do this?  I don't think this is the typical procedure for how this is done, but none are ideal so I will go along with it, I just don't want to have Minio.jl sitting around with 0.10 and then somebody tag a different 0.10 where it breaks.

ExpandingMan:
Also, FYI, I am getting my own CI/CD failures on Minio.jl right now and I have no clue why.  I have successful tests locally with multiple different binaries so I don't know what's going on.

fchorney:
>  Minio.jl sitting around with 0.10 and then somebody tag a different 0.10 where it breaks

This needs to get in ASAP for something I am working on, so I think it makes sense that this will definitely be the 0.10.0 release.

mattBrzezinski:
> > Minio.jl sitting around with 0.10 and then somebody tag a different 0.10 where it breaks
> 
> This needs to get in ASAP for something I am working on, so I think it makes sense that this will definitely be the 0.10.0 release.

Tagging this as a `0.10.0` release means we're stuck in this awkward situation when we want to do a `0.11.0` release. I'm not sure what people think, but I'm fine with moving forward to a `1.0.0` release here. The package is stable enough, and we've agreed upon being fine with doing major version bumps more frequently with breaking changes.

This would punt the need for this awkward version bump except in the case of a `2.0.0` release, but I do not foresee anyone having the time to do a major re-write of this package any time soon.

Any thoughts? @fchorney @rofinn @ExpandingMan

fchorney:
sure I'm fine with going to 1.0.0

rofinn:
> I'm not sure what people think, but I'm fine with moving forward to a 1.0.0 release here.

Yeah, I think the core behaviour hasn't changed much in the last couple years and I don't think we have a roadmap for major breaking changes. If we ever get around to doing the FilePathsBase redesign we've been discussing for a while that can just be a 2.0 release.

 https://gitlab.com/ExpandingMan/FilePaths2.jl
 
 > awkward situation when we want to do a 0.11.0 release
 
 I don't think the awkward dependency situation should guide our decision on tagging a 1.0 release though. We should probably plan to resolve the package interdependencies regardless of which release we do. We should either merge minio into a module in AWSS3 or limit AWSS3 to a test dependency for minio?

fchorney:
So in that case, we'd need `minio.jl` to allow `AWSS3.jl` v1 and then the tests here can pass and be merged in.

ExpandingMan:
# Minio.jl
I have just tagged Minio.jl 0.2.0 which nominally includes support for AWSS3.jl.  Unfortunately it seems that the minio devs are [on a warpath against single-node use cases](https://github.com/minio/minio/issues/15478) which drastically limits its usefulness for unit testing.  Besides what was mentioned in that thread, I have also noticed that minio is now rather idiosyncratic about what directory it wants to mount, even when empty; when trying to rewrite the unit tests I notice that it freaks out if you give it absolute paths like `/tmp/somepath`.  I don't think any of the AWSS3.jl tests use pre-existing directories, but there may be some subtle changes needed somewhere to get tests passing again.

Sadly at this point I would not be totally surprised if minio ultimately became totally useless for us here, but there are not many alternatives.  On the other hand, they have hit 1.0 so it'll likely be a while before they make further breaking changes.

# tagging of AWSS3.jl
I strongly recommend we wait on tagging a 1.0 of this.  In my opinion this package still has major issues that are unresolved, in particular we still lack a file path abstraction which is appropriate for all use cases.  Some of my previous comments on that can be found [here](https://github.com/JuliaCloud/AWSS3.jl/issues/254).  In my opinion we still very much need a new or completely overhauled file paths package at which point the many "gotchas" in the weirdly idiosyncratic AWSS3.jl code can be eliminated.

fchorney:
bors try

fchorney:
since I need this to go in, is everyone fine with just keeping this 0.10 and we can do a 1.0 release later.

PR #267, opened 2022-09-13
Update FilePathsBase compat

Keno:
Unfortunately FilePathsBase <= 0.9.20 is currently broken on Julia master. As such, no version of AWSS3 (which is used in PkgEval) is usable and PkgEval is broken. This just adds 0.9.20 to the list, but I don't understand why the compat list for FilePathsBase is so specific, so it might make sense to do something different.


Keno:
cc @ExpandingMan @ericphanson @omus - please take a look. This is blocking PkgEval atm, which is obviously important, so I'm trying to get this fixed if possible.

Keno:
I see that there may be more history here with #245, so I should probably leave it to you, but happy to help out if you need me to.

ExpandingMan:
In my opinion both AWSS3.jl and FilePathsBase.jl are badly in need of a major reckoning.  I think we should do whatever ugly hacks are needed to get it updated in the meantime, but I don't remember what the issue was, I don't think I was involved with the PR's that held it back.

mattBrzezinski:
Sorry, I was away on vacation these last few weeks. One day at some point this package should be re-written and made better, I know Jacob is working on [CloudStore.jl](https://github.com/JuliaServices/CloudStore.jl) to unify cloud storage solutions into one package.

I do not see myself having much time in the future to look into re-writing things here, but I'm happy to unblock people in the meantime.

ExpandingMan:
From what they told me CloudStore.jl is not supposed to have anything to do with file paths, which in my opinion are the biggest problem with this package.

The core problem is that FilePathsBase.jl was written for local file systems which are cheap to query and have true directories and doesn't work well in more general cases such as S3 in which querying the state is expensive and the API treats directories as purely implicit.  Any long term solution that did not involve discarding file paths would require it to be rewritten.

I had written [FilePaths2.jl](https://gitlab.com/ExpandingMan/FilePaths2.jl) as a prototype for a more general path type.  I was satisfied I had figured out the correct approach, however finishing the package will require a significant amount of work, mostly because now paths are essentially async in the general case and figuring out the correct semantics for updates that works for both local file systems and remote calls is challenging.

I would very much like the role of file paths in the ecosystem to be fully resolved, and @rofinn seemed to express some openness to having an overhaul like this for FilePathsBase.jl, but I was hesitant to work on finishing that without a clear consensus from the community.

ericphanson:
To solve this issue, let's just go with https://github.com/JuliaCloud/AWSS3.jl/pull/245

PR #284, opened 2023-06-05
Use OIDC to assume CI AWS role

omus:
Closes #283. Also addresses part of [this comment](https://github.com/JuliaCloud/AWSS3.jl/pull/214#issuecomment-924946837)

Sibling PR: https://github.com/JuliaCloud/AWS.jl/pull/627

PR #285, opened 2023-06-07
Use implicit keywords

omus:
Using implicit keywords which were introduced back in Julia 1.5.


omus:
bors try

omus:
I don't expect bors to work here until #284 is merged but I want to see what will happen here.

omus:
bors try-

omus:
bors try

omus:
bors try-

mattBrzezinski (approved):

PR #289, opened 2023-06-08
Use `prefix` to support policies restricting access based upon `s3:prefix`

omus:
When [S3 policies are configured to restrict access based upon prefix](https://aws.amazon.com/premiumsupport/knowledge-center/iam-s3-user-specific-folder/) some of our existence checks would fail due to not specifying a `prefix`. The PR #214 attempted to correct this but the usage of `start-after` and `prefix` within the the same query isn't supported by MinIO.

After implementing the test cases against `s3_exists` to validate the problematic behaviour it appeared that the best approach may be to dispatch on `MinioConfig`. However, since AWS.jl only depends on Minio.jl for tests and since Minio.jl depends on AWSS3.jl this fix would require PRs in both packages.

Instead, after reviewing the original implementation of `_s3_exists_dir` it occurred to me that there may be a better approach. As we all know S3 object storage is just a flat structure where each object has a key. A directory is simply a prefix of an existing object which ends in a slash. With this in mind we only care that at least one key exists with the directory prefix. See the updated docstring for `_s3_exists_dir` for more details.

I also noticed that our implementation of `mkdir(::S3Path)` will create empty directory objects for each subdirectory instead of just the deepest directory (e.g. `mkdir(p"s3://bucket/a/b/c/")` creates `s3://bucket/a/`, `s3://bucket/a/b/`, and `s3://bucket/a/b/c/` instead of just `s3://bucket/a/b/c/`). The only reason I can think of for doing this is to ensure that the parent directories will still exist once the deepest empty directory has been removed. However, since object creation doesn't work this same way I think this behaviour is just unnecessarily expensive.

<details>
<summary>Example of MinIO failure when `prefix` and `start-after` are both specified</summary>

```
AWSException: NotImplemented -- A header you provided implies functionality that is not implemented

HTTP.Exceptions.StatusError(501, "GET", "/awss3.jl.test.20230608t165213z?list-type=2&max-keys=1&delimiter=&prefix=prefix%2Fgranted%2F&start-after=prefix%2Fgranted.", HTTP.Messages.Response:
"""
HTTP/1.1 501 Not Implemented
Accept-Ranges: bytes
Content-Length: 366
Content-Security-Policy: block-all-mixed-content
Content-Type: application/xml
Server: MinIO
Strict-Transport-Security: max-age=31536000; includeSubDomains
Vary: Origin, Accept-Encoding
X-Amz-Request-Id: 1766BD4150440788
X-Content-Type-Options: nosniff
X-Xss-Protection: 1; mode=block
Date: Thu, 08 Jun 2023 16:53:18 GMT

[Message Body was streamed]""")

<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NotImplemented</Code><Message>A header you provided implies functionality that is not implemented</Message><BucketName>awss3.jl.test.20230608t165213z</BucketName><Resource>/awss3.jl.test.20230608t165213z</Resource><RequestId>1766BD4150440788</RequestId><HostId>c64d0898-6ee8-4b36-9888-d1d96029b2c0</HostId></Error>
```
</details>

Supersedes: https://github.com/JuliaCloud/AWSS3.jl/pull/214.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -259,7 +259,7 @@ end
 s3_get_meta(a...; b...) = s3_get_meta(global_aws_config(), a...; b...)
 
 function _s3_exists_file(aws::AbstractAWSConfig, bucket, path)
-    q = Dict("prefix" => path, "delimiter" => "", "max-keys" => 1)
+    q = Dict("prefix" => path, "delimiter" => "/", "max-keys" => 1)
     l = parse(S3.list_objects_v2(bucket, q; aws_config=aws))
     c = get(l, "Contents", nothing)
     c === nothing && return false
@@ -269,29 +269,23 @@ end
 """
     _s3_exists_dir(aws::AbstractAWSConfig, bucket, path)
 
-Internal function used by [`s3_exists`](@ref).
+An internal function used by [`s3_exists`](@ref).
 
-Checks whether the given directory exists.  This is a bit subtle because of how the AWS API
-handles empty directories.  Empty directories are [really just 0-byte objects](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-folders.html#create-folder)
-which are named like directories, i.e. their name has a trailing `"/"`.
+Checks if the given directory exists within the `bucket`. Since S3 uses a flat structure, as
+opposed to being hierarchical like a file system, directories are actually just a collection
+of object keys which share a common prefix. S3 implements empty directories as
+[0-byte objects](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-folders.html#create-folder)
+with keys ending with the delimiter.
 
-What this function does, given a directory name `dir/`, is check for all keys which are
-lexographically greater than `dir.`.  The reason for this is that, if `dir/` is a 0-byte
-object, checking for it directly will not reveal its existence due to some rather peculiar
-design choices on the part of the S3 developers.
-
-In all such cases, if the directory exists it will be seen in the *first* item returned from
-`S3.list_objects_v2`: for empty directories this is because using `start-after` explicitly
-excludes `dir.` itself and `dir/` is next; for directories with actual keys, it is
-guaranteed that the first contained key will start with the directory name.
+It is possible to create non 0-byte objects with a key ending in the delimiter
+(e.g. `s3_put(bucket, "abomination/", "If I cannot inspire love, I will cause fear!")`)
+which the AWS console interprets as the directory "abmonination" containing the object "/".
 """
 function _s3_exists_dir(aws::AbstractAWSConfig, bucket, path)
-    a = chop(string(path)) * "."
-    q = Dict("prefix" => path, "delimiter" => "", "max-keys" => 1, "start-after" => a)
-    l = parse(S3.list_objects_v2(bucket, q; aws_config=aws))
-    c = get(l, "Contents", nothing)
-    c === nothing && return false
-    return startswith(get(c, "Key", ""), path)
+    endswith(path, '/') || throw(ArgumentError("S3 directories must end with '/': $path"))
+    q = Dict("prefix" => path, "delimiter" => "/", "max-keys" => 1)
+    r = parse(S3.list_objects_v2(bucket, q; aws_config=aws))
+    return get(r, "KeyCount", "0") != "0"
 end
 
 """
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -288,7 +288,7 @@ the directory name.
 """
 function _s3_exists_dir(aws::AbstractAWSConfig, bucket, path)
     a = chop(string(path)) * "."
-    q = Dict("delimiter" => "", "max-keys" => 1, "start-after" => a, "prefix" => path)
+    q = Dict("prefix" => path, "delimiter" => "", "max-keys" => 1, "start-after" => a)
     l = parse(S3.list_objects_v2(bucket, q; aws_config=aws))
     c = get(l, "Contents", nothing)
     c === nothing && return false
@@ -753,15 +753,18 @@ function s3_list_objects(
         token = ""
 
         while more
-            q = Dict{String,String}()
+            q = Dict{String,String}(
+                "prefix" => path_prefix
+            )
+
             for (name, v) in [
-                ("prefix", path_prefix),
                 ("delimiter", delimiter),
                 ("start-after", start_after),
                 ("continuation-token", token),
             ]
                 isempty(v) || (q[name] = v)
             end
+
             if max_items !== nothing
                 # Note: AWS seems to only return up to 1000 items
                 q["max-keys"] = string(max_items - num_objects)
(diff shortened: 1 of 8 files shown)


omus:
bors try

src/AWSS3.jl
```diff
@@ -258,7 +258,7 @@ end
 s3_get_meta(a...; b...) = s3_get_meta(global_aws_config(), a...; b...)
 
 function _s3_exists_file(aws::AbstractAWSConfig, bucket, path)
-    q = Dict("prefix" => path, "delimiter" => "", "max-keys" => 1)
+    q = Dict("prefix" => path, "delimiter" => "/", "max-keys" => 1)
```
iamed2, line 261: What does this do here?

omus: This is required by https://github.com/JuliaCloud/AWSS3.jl/pull/289/files#diff-0622bb94c91db091998eec0f64f90430f91c0440b8a0e19fd71cdd45c82be123R266. As the [blog post](https://repost.aws/knowledge-center/iam-s3-user-specific-folder) I based the tests off uses this I think we should update the delimiter used here (tests do fail without this)

omus: I'll double check that there are tests in place that require this.

omus:
Looks like the `@test_throws` syntax I'm using isn't supported in Julia 1.6. I'll look into that.

src/s3path.jl
```diff
@@ -418,12 +417,13 @@ Base.ismount(fp::S3Path) = false
 """
     mkdir(fp::S3Path; recursive=false, exist_ok=false)
 
-Create an empty directory at the S3 path `fp`.  If `recursive`, this will create any previously non-existent
-directories which would contain `fp`.  An error will be thrown if an object exists at `fp` unless `exist_ok`.
-
-Note that empty directories in S3 are actually 0-byte objects with the naming convention of a directory.
+Create an empty directory within an existing bucket for the S3 path `fp`. If `recursive`,
+this will create any previously non-existent directories which would contain `fp`. An error
+will be thrown if an object exists at `fp` unless `exist_ok`.
```
kleinschmidt, line 422: this would be pretty breaking but IMO the "recursive" case should actually be handled by `mkpath` instead of `mkdir`.

src/AWSS3.jl
kleinschmidt: just to check my own understanding: when you specify `delimiter="/"`, this has the effect of "collapsing" all subdirectories below `prefix` to a single entry, effectively allowing the user to see _which_ subdirectories exist below `path`, but not get any information about their contents.  is that right?

kleinschmidt (commented):
a bit out of my depth here.  I _think_ I follow what's going on but I'd be more comfortable if someone with more AWS experience had a look.

the added tests are consistent with my understanding though, so this looks good as far as I can tell.

src/s3path.jl
omus: I completely agree. I was going to do that as part of this PR but decided to keep that for another PR due to the breaking nature of the change.

src/AWSS3.jl
omus: > just to check my own understanding: when you specify `delimiter="/"`, this has the effect of "collapsing" all subdirectories below `prefix` to a single entry, effectively allowing the user to see _which_ subdirectories exist below `path`, but not get any information about their contents. is that right?

Your understanding is correct. Here's the documentation on the subject: https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html

Here's a concrete example:
```sh
❯ aws-vault exec ReadObject --region us-east-1 -- aws s3api list-objects-v2 --bucket awss3.jl.test.20230609t183606z --prefix prefix/
- Contents:
  - ETag: '"fbfa50731929244e205bbf7dc3737bcf"'
    Key: prefix/denied/secrets/top-secret
    LastModified: '2023-06-09T18:36:38+00:00'
    Size: 21
    StorageClass: STANDARD
  - ETag: '"5d41402abc4b2a76b9719d911017c592"'
    Key: prefix/granted/file
    LastModified: '2023-06-09T18:36:38+00:00'
    Size: 5
    StorageClass: STANDARD
  EncodingType: url
  IsTruncated: false
  KeyCount: 2
  MaxKeys: 1000
  Name: awss3.jl.test.20230609t183606z
  Prefix: prefix/

❯ aws-vault exec ReadObject --region us-east-1 -- aws s3api list-objects-v2 --bucket awss3.jl.test.20230609t183606z --prefix prefix/ --delimiter /
- CommonPrefixes:
  - Prefix: prefix/denied/
  - Prefix: prefix/granted/
  Delimiter: /
  EncodingType: url
  IsTruncated: false
  KeyCount: 2
  MaxKeys: 1000
  Name: awss3.jl.test.20230609t183606z
  Prefix: prefix/
```
The reason though we need to include the delimiter in these queries is that our AWS policies allowing us to list bucket content are conditional on this `s3:delimiter: ["/"]`. If you do not pass in this delimiter than the policy is violated and you cannot list the content. As `s3_exists` was already using `endswith(path, '/')` in it's logic before this PR we've already chosen using "/" as our delimiter so this change just allows us to use policies that require the delimiter to be set.

omus: Since a change to the delimiter in this code may violate an existing AWS IAM policy using `s3:delimiter: [""]` I'll mark this change as breaking.

omus:
Ugh, the AWSS3 tests cannot run as Minio hasn't declared compatibility with the latest version:
```
 ERROR: LoadError: Unsatisfiable requirements detected for package AWSS3 [1c724243]:
 AWSS3 [1c724243] log:
 ├─possible versions are: 0.11.0 or uninstalled
 ├─AWSS3 [1c724243] is fixed to version 0.11.0
 └─found to have no compatible versions left with Minio [4281f0d9]
   └─Minio [4281f0d9] log:
     ├─possible versions are: 0.1.0-0.2.0 or uninstalled
     └─restricted to versions 0.1-0.2 by an explicit requirement, leaving only versions: 0.1.0-0.2.0
```
I'll revert the version bump here just to show the tests work

omus:
bors try

omus:
Made a PR to Minio.jl to state that it's compatible with these changes: https://gitlab.com/ExpandingMan/Minio.jl/-/merge_requests/4

src/s3path.jl
kleinschmidt: eh, if we're gonna make this breaking anyway due to the delimiter change, might as well do it all at once here.  or at least add the `mkpath` method and deprecate the recursive kwarg

src/AWSS3.jl
```diff
@@ -759,15 +751,16 @@ function s3_list_objects(
         token = ""
 
         while more
-            q = Dict{String,String}()
+            q = Dict{String,String}("prefix" => path_prefix)
```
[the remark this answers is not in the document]
mattBrzezinski: Right... I always forget that `isempty(nothing)` doesn't work.

omus:
bors try

kleinschmidt (approved):
okay, with the changelog and some offline discussions for how this interact with our policies, I'm comfortable with this!

PR #287, opened 2023-06-07
Document used permissions and restrict permissions used in tests

omus:
While working on #284 it became evident that we haven't been keeping track of what AWS API calls are used when calling our various exported functions. I started documenting which API calls are used and the required permissions and found myself wanting to update the tests to limit the permissions required for each testset. This PR is the result of those ideas and should allow us to write better test for things like: https://github.com/JuliaCloud/AWSS3.jl/pull/214#issuecomment-924943077.

I'll mention that I didn't document every single function's permissions as part of this PR but I did try to document anything I needed to validate as part of tests. The remaining functions can be done as a follow up.

PR #291, opened 2023-06-12
Allow tags to assume OIDC role AWSS3.jl

omus:
Follow up to #284. When the CI is triggered from a tag the job would be unable to assume the AWSS3.jl role due to a missing trust policy. This was noticed on the v0.10.4 tag and originally thought to be from the PR https://github.com/JuliaCloud/AWSS3.jl/pull/287#issuecomment-1587488362


omus:
bors try

Issue #213, opened 2021-09-22
Segmentation Fault s3_get


mahiki:
I'm able to do AWSS3.read of s3 path to object, but in the same project I'm able to repeatedly hit a seg fault when using `AWS.s3_get`. Here is the error:

```jl
using AWSS3, AWS
aws = global_aws_config(; region = "us-east-1")
# AWSConfig((AKIA??????????H6, I5j..., 146138512-12-31T23:59:59), "us-east-1", "json")

AWSS3.s3_get(aws, "test-data", "test/test-37/s3-walkpath", "df-data-1.txt")
# zsh: segmentation fault  julia --project=. --startup-file=no

# using the correct number of arguments had no problem
AWSS3.s3_get(aws, "test-data", "test/test-37/s3-walkpath/df-data-1.txt")

# this is correct object contents:
#=
"column_1\tamount\titem\tday\textra\n0\t19.0\tB001\t2021-01-01\tHello newfile\n1\t11.0\tB001\t2021-01-01\tHello newfile\n0\t35.5\tB020\t2112-12-12\tHello newfile\n1\t32.5\tB020\t2020-10-20\tHello newfile\n0\t5.99\tBX00\t2021-05-04\tHello newfile\n1\t5.99\tBX00\t1984-07-04\tHello newfile\n"
=#
```

I mistakenly passed the "bucket", "path", "filename" instead of "bucket", "path/filename" arguments, because I was thinking of the AWSS3.s3_get_file() call which takes "filename" as a separate argument.

Here is the project configuration:

```jl
pkg> st
  [fbe9abb3] AWS v1.61.0
  [1c724243] AWSS3 v0.9.2

```

-------
I also find I was confused by the documentation of s3_get_file, resulted in me chasing around the error stacktrace. The documentation is clear, once I realize my mistake, but `target_filename`  or `local_filename` (as in the awscli documentation) might be better than `filename`?

```jl
help?> AWSS3.s3_get_file
  s3_get_file([::AbstractAWSConfig], bucket, path, filename; [version=], kwargs...)
```

It wasn't immediately obvious "filename" is the *target* filename on the local system, and "path" is key to the object in s3.

The AWS CLI doc looks like:
```
aws s3 cp help
SYNOPSIS
            cp
          <LocalPath> <S3Uri> or <S3Uri> <LocalPath> or <S3Uri> <S3Uri>
```

mahiki:
[...]

PR #298, opened 2023-06-28
Update variable arg lists to named arg lists

hannahilea:
Address #213 by updating `a...` to the list of named arguments for functions with 3 or fewer arguments.

Somewhat pedantic, but it is very easy for folks new to this library to get confusing error messages back, and this will helpfully lead to clearer error messages.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -558,8 +558,8 @@ function s3_enable_versioning(aws::AbstractAWSConfig, bucket, status="Enabled"; 
     return parse(r)
 end
 
-function s3_enable_versioning(bucket; kwargs...)
-    return s3_enable_versioning(global_aws_config(), bucket; kwargs...)
+function s3_enable_versioning(args...; kwargs...)
+    return s3_enable_versioning(global_aws_config(), args...; kwargs...)
 end
 
 """
@@ -793,8 +793,9 @@ function s3_list_objects(
         end
     end
 end
-function s3_list_objects(bucket, path_prefix=""; kwargs...)
-    return s3_list_objects(global_aws_config(), bucket, path_prefix; kwargs...)
+
+function s3_list_objects(args...; kwargs...)
+    return s3_list_objects(global_aws_config(), args...; kwargs...)
 end
 
 """
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -204,7 +204,9 @@ function s3_get_file(
     end
 end
 
-s3_get_file(bucket, path, filename; kwargs...) = s3_get_file(global_aws_config(), bucket, path, filename; kwargs...)
+function s3_get_file(bucket, path, filename; kwargs...)
+    return s3_get_file(global_aws_config(), bucket, path, filename; kwargs...)
+end
 
 function s3_get_file(
     aws::AbstractAWSConfig,
@@ -257,7 +259,9 @@ function s3_get_meta(
     return Dict(r.headers)
 end
 
-s3_get_meta(bucket, path; kwargs...) = s3_get_meta(global_aws_config(), bucket, path; kwargs...)
+function s3_get_meta(bucket, path; kwargs...)
+    return s3_get_meta(global_aws_config(), bucket, path; kwargs...)
+end
 
 function _s3_exists_file(aws::AbstractAWSConfig, bucket, path)
     q = Dict("prefix" => path, "delimiter" => "/", "max-keys" => 1)
@@ -518,7 +522,9 @@ function s3_put_cors(aws::AbstractAWSConfig, bucket, cors_config; kwargs...)
     return parse(S3.put_bucket_cors(bucket, cors_config; aws_config=aws, kwargs...))
 end
 
-s3_put_cors(bucket, cors_config; kwargs...) = s3_put_cors(AWS.global_aws_config(), bucket, cors_config; kwargs...)
+function s3_put_cors(bucket, cors_config; kwargs...)
+    return s3_put_cors(AWS.global_aws_config(), bucket, cors_config; kwargs...)
+end
 
 """
     s3_enable_versioning([::AbstractAWSConfig], bucket, [status]; kwargs...)
@@ -598,8 +604,12 @@ function s3_put_tags(aws::AbstractAWSConfig, bucket, tags::SSDict; kwargs...)
     return s3_put_tags(aws, bucket, "", tags; kwargs...)
 end
 
-s3_put_tags(bucket, path, tags; kwargs...) = s3_put_tags(global_aws_config(), bucket, path, tags; kwargs...)
-s3_put_tags(bucket, tags; kwargs...) = s3_put_tags(global_aws_config(), bucket, tags; kwargs...)
+function s3_put_tags(bucket, path, tags; kwargs...)
+    return s3_put_tags(global_aws_config(), bucket, path, tags; kwargs...)
+end
+function s3_put_tags(bucket, tags; kwargs...)
+    return s3_put_tags(global_aws_config(), bucket, tags; kwargs...)
+end
 
 """
     s3_get_tags([::AbstractAWSConfig], bucket, [path]; kwargs...)
@@ -647,7 +657,9 @@ function s3_get_tags(aws::AbstractAWSConfig, bucket, path=""; kwargs...)
     end
 end
 
-s3_get_tags(bucket, path=""; kwargs...) = s3_get_tags(global_aws_config(), bucket, path; kwargs...)
+function s3_get_tags(bucket, path=""; kwargs...)
+    return s3_get_tags(global_aws_config(), bucket, path; kwargs...)
+end
 
 """
     s3_delete_tags([::AbstractAWSConfig], bucket, [path])
@@ -678,7 +690,9 @@ function s3_delete_tags(aws::AbstractAWSConfig, bucket, path=""; kwargs...)
     return parse(r)
 end
 
-s3_delete_tags(bucket, path=""; kwargs...) = s3_delete_tags(global_aws_config(), bucket, path; kwargs...)
+function s3_delete_tags(bucket, path=""; kwargs...)
+    return s3_delete_tags(global_aws_config(), bucket, path; kwargs...)
+end
 
 """
     s3_delete_bucket([::AbstractAWSConfig], "bucket"; kwargs...)
@@ -784,7 +798,9 @@ function s3_list_objects(
         end
     end
 end
-s3_list_objects(bucket, path_prefix=""; kw...) = s3_list_objects(global_aws_config(), bucket, path_prefix; kw...)
+function s3_list_objects(bucket, path_prefix=""; kw...)
+    return s3_list_objects(global_aws_config(), bucket, path_prefix; kw...)
+end
 
 """
     s3_directory_stat([::AbstractAWSConfig], bucket, path)
@@ -820,7 +836,9 @@ function s3_list_keys(aws::AbstractAWSConfig, bucket, path_prefix=""; kwargs...)
     return (o["Key"] for o in s3_list_objects(aws, bucket, path_prefix; kwargs...))
 end
 
-s3_list_keys(bucket, path_prefix=""; kwargs...) = s3_list_keys(global_aws_config(), bucket, path_prefix; kwargs...)
+function s3_list_keys(bucket, path_prefix=""; kwargs...)
+    return s3_list_keys(global_aws_config(), bucket, path_prefix; kwargs...)
+end
 
 """
     s3_list_versions([::AbstractAWSConfig], bucket, [path_prefix]; kwargs...)
@@ -866,7 +884,9 @@ function s3_list_versions(aws::AbstractAWSConfig, bucket, path_prefix=""; kwargs
     return versions
 end
 
-s3_list_versions(bucket, path_prefix=""; kwargs...) = s3_list_versions(global_aws_config(), bucket, path_prefix; kwargs...)
+function s3_list_versions(bucket, path_prefix=""; kwargs...)
+    return s3_list_versions(global_aws_config(), bucket, path_prefix; kwargs...)
+end
 
 """
     s3_purge_versions([::AbstractAWSConfig], bucket, [path_prefix [, pattern]]; kwargs...)
@@ -905,7 +925,9 @@ function s3_purge_versions(
     end
 end
 
-s3_purge_versions(bucket, path_prefix="", pattern=""; kwargs...) = s3_purge_versions(global_aws_config(), bucket, path_prefix, pattern; kwargs...)
+function s3_purge_versions(bucket, path_prefix="", pattern=""; kwargs...)
+    return s3_purge_versions(global_aws_config(), bucket, path_prefix, pattern; kwargs...)
+end
 
 """
     s3_put([::AbstractAWSConfig], bucket, path, data, data_type="", encoding="";
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -558,7 +558,7 @@ function s3_enable_versioning(aws::AbstractAWSConfig, bucket, status="Enabled"; 
     return parse(r)
 end
 
-s3_enable_versioning(a; kwargs...) = s3_enable_versioning(global_aws_config(), a; kwargs...)
+s3_enable_versioning(bucket; kwargs...) = s3_enable_versioning(global_aws_config(), bucket; kwargs...)
 
 """
     s3_put_tags([::AbstractAWSConfig], bucket, [path], tags::Dict; kwargs...)
@@ -798,8 +798,8 @@ function s3_list_objects(
         end
     end
 end
-function s3_list_objects(bucket, path_prefix=""; kw...)
-    return s3_list_objects(global_aws_config(), bucket, path_prefix; kw...)
+function s3_list_objects(bucket, path_prefix=""; kwargs...)
+    return s3_list_objects(global_aws_config(), bucket, path_prefix; kwargs...)
 end
 
 """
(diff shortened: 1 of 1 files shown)


src/AWSS3.jl
```diff
@@ -552,7 +558,7 @@ function s3_enable_versioning(aws::AbstractAWSConfig, bucket, status="Enabled";
     return parse(r)
 end
 
-s3_enable_versioning(a; b...) = s3_enable_versioning(global_aws_config(), a; b...)
+s3_enable_versioning(a; kwargs...) = s3_enable_versioning(global_aws_config(), a; kwargs...)
```
mattBrzezinski, line 561:
```suggestion
s3_enable_versioning(bucket; kwargs...) = s3_enable_versioning(global_aws_config(), bucket; kwargs...)
```

src/AWSS3.jl
```diff
@@ -598,7 +604,12 @@ function s3_put_tags(aws::AbstractAWSConfig, bucket, tags::SSDict; kwargs...)
     return s3_put_tags(aws, bucket, "", tags; kwargs...)
 end
 
-s3_put_tags(a...) = s3_put_tags(global_aws_config(), a...)
+function s3_put_tags(bucket, path, tags; kwargs...)
+    return s3_put_tags(global_aws_config(), bucket, path, tags; kwargs...)
+end
+function s3_put_tags(bucket, tags; kwargs...)
+    return s3_put_tags(global_aws_config(), bucket, tags; kwargs...)
```
mattBrzezinski, line 611: This wouldn't work as `path` has no default value?

src/AWSS3.jl
```diff
@@ -783,7 +798,9 @@ function s3_list_objects(
         end
     end
 end
-s3_list_objects(a...; kw...) = s3_list_objects(global_aws_config(), a...; kw...)
+function s3_list_objects(bucket, path_prefix=""; kw...)
+    return s3_list_objects(global_aws_config(), bucket, path_prefix; kw...)
+end
```
mattBrzezinski, line 803: Just for consistency,

```suggestion
function s3_list_objects(bucket, path_prefix=""; kwargs...)
    return s3_list_objects(global_aws_config(), bucket, path_prefix; kwargs...)
end
```

src/AWSS3.jl
omus, line 561: Looks like this should be updated to `args...`:
```suggestion
s3_enable_versioning(args...; kwargs...) = s3_enable_versioning(global_aws_config(), args...; kwargs...)
```
The `bucket` and `status` function cannot be called without also passing in an `AbstractAWSConfig`:
```julia
julia> methods(s3_enable_versioning)
# 3 methods for generic function "s3_enable_versioning":
[1] s3_enable_versioning(aws::AWS.AbstractAWSConfig, bucket) in AWSS3 at /Users/cvogt/.julia/dev/AWSS3/src/AWSS3.jl:537
[2] s3_enable_versioning(aws::AWS.AbstractAWSConfig, bucket, status; kwargs...) in AWSS3 at /Users/cvogt/.julia/dev/AWSS3/src/AWSS3.jl:537
[3] s3_enable_versioning(a; b...) in AWSS3 at /Users/cvogt/.julia/dev/AWSS3/src/AWSS3.jl:555
```

src/AWSS3.jl
```diff
@@ -598,7 +604,12 @@ function s3_put_tags(aws::AbstractAWSConfig, bucket, tags::SSDict; kwargs...)
     return s3_put_tags(aws, bucket, "", tags; kwargs...)
 end
 
-s3_put_tags(a...) = s3_put_tags(global_aws_config(), a...)
+function s3_put_tags(bucket, path, tags; kwargs...)
+    return s3_put_tags(global_aws_config(), bucket, path, tags; kwargs...)
+end
+function s3_put_tags(bucket, tags; kwargs...)
+    return s3_put_tags(global_aws_config(), bucket, tags; kwargs...)
+end
```
omus, line 612: Makes more sense to use `args...` here to avoid having to specify additional methods:
```suggestion
s3_put_tags(args...; kwargs...) = s3_put_tags(global_aws_config(), args...; kwargs...)
```

src/AWSS3.jl
```diff
@@ -646,7 +657,9 @@ function s3_get_tags(aws::AbstractAWSConfig, bucket, path=""; kwargs...)
     end
 end
 
-s3_get_tags(a...; b...) = s3_get_tags(global_aws_config(), a...; b...)
+function s3_get_tags(bucket, path=""; kwargs...)
+    return s3_get_tags(global_aws_config(), bucket, path; kwargs...)
+end
```
omus, line 662: I see you're specifying `path=""` to avoid defining two methods. Usually in this case I'd use `args...` as it avoids having to set the default value twice leading to the potential for them to become different over time. Alternatively, you could use a `const` and reference it in both places:
```suggestion
s3_get_tags(args...; kwargs...) = s3_get_tags(global_aws_config(), args...; kwargs...)
```

src/AWSS3.jl
```diff
@@ -981,7 +1004,7 @@ function s3_put(
     return parse_response ? parse(response) : response
 end
 
-s3_put(a...; b...) = s3_put(global_aws_config(), a...; b...)
+s3_put(args...; kwargs...) = s3_put(global_aws_config(), args...; kwargs...)
```
mattBrzezinski, line 1007:
```suggestion
function s3_put(bucket, path, data::Union{String,Vector{UInt8}}, data_type="", encoding=""; kwargs...
  return s3_put(global_aws_config(), bucket, path, data, data_type, encoding; kwargs...)
end
```

src/AWSS3.jl
```diff
@@ -677,7 +690,9 @@ function s3_delete_tags(aws::AbstractAWSConfig, bucket, path=""; kwargs...)
     return parse(r)
 end
 
-s3_delete_tags(a...; b...) = s3_delete_tags(global_aws_config(), a...; b...)
+function s3_delete_tags(bucket, path=""; kwargs...)
+    return s3_delete_tags(global_aws_config(), bucket, path; kwargs...)
+end
```
omus, line 695: Same issue as above:
```suggestion
s3_delete_tags(args...; kwargs...) = s3_delete_tags(global_aws_config(), args...; kwargs...)
```

src/AWSS3.jl
```diff
@@ -1259,7 +1282,7 @@ function s3_sign_url(
     end
 end
 
-s3_sign_url(a...; b...) = s3_sign_url(global_aws_config(), a...; b...)
+s3_sign_url(args...; kwargs...) = s3_sign_url(global_aws_config(), args...; kwargs...)
```
mattBrzezinski, line 1285:
```suggestion
function s3_sign_url(bucket, path, seconds=3600; kwargs...)
  return s3_sign_url(global_aws_config(), bucket, path, seconds; kwargs...)
end
```

src/AWSS3.jl
```diff
@@ -698,7 +713,7 @@ See also [`AWSS3.s3_nuke_bucket`](@ref).
 function s3_delete_bucket(aws::AbstractAWSConfig, bucket; kwargs...)
     return parse(S3.delete_bucket(bucket; aws_config=aws, kwargs...))
 end
-s3_delete_bucket(a; b...) = s3_delete_bucket(global_aws_config(), a; b...)
+s3_delete_bucket(a; kwargs...) = s3_delete_bucket(global_aws_config(), a; kwargs...)
```
omus, line 716: May as well name the argument:
```suggestion
function s3_delete_bucket(bucket; kwargs...)
    return s3_delete_bucket(global_aws_config(), bucket; kwargs...)
end
```

mattBrzezinski (approved):

src/AWSS3.jl
omus, line 803: Avoid redefining default values:
```suggestion
function s3_list_objects(args...; kwargs...)
    return s3_list_objects(global_aws_config(), args...; kwargs...)
end
```

src/AWSS3.jl
```diff
@@ -819,7 +836,9 @@ function s3_list_keys(aws::AbstractAWSConfig, bucket, path_prefix=""; kwargs...)
     return (o["Key"] for o in s3_list_objects(aws, bucket, path_prefix; kwargs...))
 end
 
-s3_list_keys(a...; b...) = s3_list_keys(global_aws_config(), a...; b...)
+function s3_list_keys(bucket, path_prefix=""; kwargs...)
+    return s3_list_keys(global_aws_config(), bucket, path_prefix; kwargs...)
+end
```
omus, line 841: Avoid redefining default values:
```suggestion
s3_list_keys(args...; kwargs...) = s3_list_keys(global_aws_config(), args...; kwargs...)
```

src/AWSS3.jl
```diff
@@ -865,7 +884,9 @@ function s3_list_versions(aws::AbstractAWSConfig, bucket, path_prefix=""; kwargs
     return versions
 end
 
-s3_list_versions(a...; b...) = s3_list_versions(global_aws_config(), a...; b...)
+function s3_list_versions(bucket, path_prefix=""; kwargs...)
+    return s3_list_versions(global_aws_config(), bucket, path_prefix; kwargs...)
+end
```
omus, line 889: Avoid redefining default values:
```suggestion
function s3_list_versions(args...; kwargs...)
    return s3_list_versions(global_aws_config(), args...; kwargs...)
end
```

src/AWSS3.jl
```diff
@@ -904,7 +925,9 @@ function s3_purge_versions(
     end
 end
 
-s3_purge_versions(a...; b...) = s3_purge_versions(global_aws_config(), a...; b...)
+function s3_purge_versions(bucket, path_prefix="", pattern=""; kwargs...)
+    return s3_purge_versions(global_aws_config(), bucket, path_prefix, pattern; kwargs...)
+end
```
omus, line 930: Avoid redefining default values:
```suggestion
function s3_purge_versions(args...; kwargs...)
    return s3_purge_versions(global_aws_config(), args...; kwargs...)
end
```

omus (commented):
I overall agree with the direction this PR is going in. Unfortunately, moving away from using `args...` to a fixed list of arguments would have to be considered a breaking change as any package extending on of these methods may rely on the `args...` version to fill in the default AWS config. 

Additionally, I prefer not to have to redefine the same default values as it can lead to issues where they become out of sync. I think `args...` is probably a better option here but we could also use a `const` to address this particular problem.

src/AWSS3.jl
hannahilea: no longer valid (followed the below @omus style)

src/AWSS3.jl
hannahilea: no longer valid (followed the below @omus style)

src/AWSS3.jl
```diff
@@ -179,7 +179,7 @@ function s3_get(
     end
 end
 
-s3_get(a...; b...) = s3_get(global_aws_config(), a...; b...)
+s3_get(bucket, path; kwargs...) = s3_get(global_aws_config(), bucket, path; kwargs...)
```
omus, line 182: Last thing to deal with: as we're moving from `args...` to specifically support a fixed amount of args this will need to be marked as breaking. Did you want to make this a breaking change?

hannahilea: :lol-cry: this was the originating issue: #213. The point in making this switch in the first place was to make it less likely for newer users to get sensical error messages.... so while this seems like a silly issue to make a breaking release over, without making this change then this whole PR was kinda pointless. as a maintainer, lmk which you'd prefer---we can also ditch this PR outright!

omus: I should have read through the original issue. My bad. I think there may be a non-breaking alternative here. I'll dig into it

omus: Okay, I've cooked up a non-breaking pattern:
```julia
julia> using AWS

julia> foo(::AbstractAWSConfig, bucket, key) = (bucket, key)
foo (generic function with 1 method)

julia> foo(args...) = foo(global_aws_config(), args...)
foo (generic function with 2 methods)

julia> foo("bucket", "key")
("bucket", "key")

julia> foo("bucket", "key", 0)
ERROR: StackOverflowError:

julia> foo(aws::AbstractAWSConfig, args...) = throw(MethodError(foo, Tuple{typeof(aws), typeof(args).types...}))
foo (generic function with 3 methods)

julia> foo("bucket", "key", 0)
ERROR: MethodError: no method matching foo(::AWSConfig, ::String, ::String, ::Int64)
Closest candidates are:
  foo(::AbstractAWSConfig, ::Any...) at REPL[6]:1
  foo(::Any...) at REPL[3]:1
  foo(::AbstractAWSConfig, ::Any, ::Any) at REPL[2]:1
Stacktrace:
 [1] foo(::AWSConfig, ::String, ::String, ::Int64)
   @ Main ./REPL[6]:1
 [2] foo(::String, ::String, ::Int64)
   @ Main ./REPL[3]:1
 [3] top-level scope
   @ REPL[7]:1

julia> foo("bucket", "key")
("bucket", "key")
```

omus, line 182:
```suggestion
s3_get(args...; kwargs...) = s3_get(global_aws_config(), args...; kwargs...)
function s3_get(aws::AbstractAWSConfig, args...; kwargs...)
    throw(MethodError(s3_get, Tuple{typeof(aws), typeof(args).types...}))
end
```

src/AWSS3.jl
```diff
@@ -204,7 +204,9 @@ function s3_get_file(
     end
 end
 
-s3_get_file(a...; b...) = s3_get_file(global_aws_config(), a...; b...)
+function s3_get_file(bucket, path, filename; kwargs...)
+    return s3_get_file(global_aws_config(), bucket, path, filename; kwargs...)
+end
```
omus, line 209:
```suggestion
function s3_get_file(args...; kwargs...)
    return s3_get_file(global_aws_config(), args...; kwargs...)
end
function s3_get_file(aws::AbstractAWSConfig, args...; kwargs...)
    throw(MethodError(s3_get_file, Tuple{typeof(aws), typeof(args).types...}))
end
```

src/AWSS3.jl
```diff
@@ -257,7 +259,9 @@ function s3_get_meta(
     return Dict(r.headers)
 end
 
-s3_get_meta(a...; b...) = s3_get_meta(global_aws_config(), a...; b...)
+function s3_get_meta(bucket, path; kwargs...)
+    return s3_get_meta(global_aws_config(), bucket, path; kwargs...)
+end
```
omus, line 264:
```suggestion
function s3_get_meta(args...; kwargs...)
    return s3_get_meta(global_aws_config(), args...; kwargs...)
end
function s3_get_meta(aws::AbstractAWSConfig, args...; kwargs...)
    throw(MethodError(s3_get_meta, Tuple{typeof(aws), typeof(args).types...}))
end
```

omus (commented):
Couple of examples on how to fix this. I'll cook up a macro

omus:
```julia
macro optional_aws_config(expr)
    return esc(_optional_aws_config(expr))
end

function _optional_aws_config(expr::Expr)
    def = splitdef(expr)

    if haskey(def, :name)
        func = def[:name]
    else
        throw(ArgumentError("Function definition must be a named function"))
    end

    if length(def[:args]) > 1 && def[:args][1] isa Expr && def[:args][1].head === :(::)
        args = def[:args][1].args
        T = if 1 <= length(args) <= 2
            args[end]
        else
            throw(ArgumentError("Unhandled function argument expression: $(def[:args][1])"))
        end

        if T !== :AbstractAWSConfig
            throw(ArgumentError("First argument of function is required to be of type `AbstractAWSConfig`"))
        end
    else
        throw(ArgumentError("First argument of function is required to contain a type assertion"))
    end

    return quote
        $expr
        $func(args...; kwargs...) = $func(global_aws_config(), args...; kwargs...)
        function $func(aws::AbstractAWSConfig, args...; kwargs...)
            throw(MethodError($func, Tuple{typeof(aws), typeof(args).types...}))
        end
    end
end
```

Can be used like this:
```julia
@optional_aws_config function s3_get_file(
    aws::AbstractAWSConfig,
    bucket,
    path,
    filename;
    version::AbstractS3Version=nothing,
    kwargs...,
)
    ...
end
```

omus (changes requested):
Shouldn't be merged without addressing the potentially breaking behaviour

hannahilea:
low priority, so (as discussed w/ @omus) abandoning for the time being.

nhz2:
I just tried out this library and got the following weird error message.

```julia
julia> s3_put(cfg, "my.bucket", "foo.txt", codeunits("hello world"))
ERROR: Can't find AWS credentials!

Stacktrace:
 [1] AWSCredentials(; profile::Nothing, throw_cred_error::Bool)
   @ AWS ~/.julia/packages/AWS/PRwK5/src/AWSCredentials.jl:145
 [2] AWSCredentials
   @ ~/.julia/packages/AWS/PRwK5/src/AWSCredentials.jl:113 [inlined]
 [3] AWS.AWSConfig()
   @ AWS ~/.julia/packages/AWS/PRwK5/src/AWSConfig.jl:32
 [4] #global_aws_config#91
   @ ~/.julia/packages/AWS/PRwK5/src/AWS.jl:80 [inlined]
 [5] global_aws_config
   @ ~/.julia/packages/AWS/PRwK5/src/AWS.jl:78 [inlined]
 [6] s3_put(::MinioConfig, ::Vararg{Any}; b::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ AWSS3 ~/.julia/packages/AWSS3/8cxdr/src/AWSS3.jl:1041
 [7] s3_put(::MinioConfig, ::String, ::String, ::Base.CodeUnits{UInt8, String})
   @ AWSS3 ~/.julia/packages/AWSS3/8cxdr/src/AWSS3.jl:1041
 [8] top-level scope
   @ REPL[15]:1
```

I'm pretty sure this error is also caused by the strange `::Vararg{Any}` methods and has nothing to do with AWS credentials.

PR #299, opened 2023-06-28
Add `s3_nuke_object`

hannahilea:
The existing `s3_delete` allows the deletion of either the most recent or a specific, specified version of an object; this new ~`s3_delete_all_versions`~ `s3_nuke_object` deletes _all_ versions of that object.

Upstreamed from internal code originally written by @kleinschmidt

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -411,6 +411,37 @@ end
 s3_delete(bucket, path; kwargs...) = s3_delete(global_aws_config(), bucket, path; kwargs...)
 
 """
+    s3_delete_all_versions([::AbstractAWSConfig], bucket, path; kwargs...)
+
+Deletes all versions of an object from a bucket; forwards all `kwargs` to internal
+[`s3_delete`](@ref) call.
+
+# API Calls
+
+- [`DeleteObject`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html)
+- [`ListObjectVersions`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html)
+
+# Permissions
+
+- [`s3:DeleteObject`](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html#amazons3-DeleteObject)
+- [`s3:ListBucketVersions`](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html#amazons3-ListBucketVersions)
+"""
+function s3_delete_all_versions(aws::AbstractAWSConfig, bucket, path; kwargs...)
+    versions = [x["VersionId"] for x in s3_list_versions(bucket, path)]
+    for version in versions
+        try
+            s3_delete(path.bucket, path.key; version, kwargs...)
+        catch e
+            @error "Error deleting version $(version) of $(path)\n\n$(e)"
+        end
+    end
+    return nothing
+end
+
+s3_delete_all_versions(bucket, path; kwargs...) = s3_delete_all_versions(global_aws_config(), bucket, path; kwargs...)
+
+
+"""
     s3_copy([::AbstractAWSConfig], bucket, path; to_bucket=bucket, to_path=path, kwargs...)
 
 [PUT Object - Copy](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html)
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -15,7 +15,6 @@ export S3Path,
     s3_get_file,
     s3_exists,
     s3_delete,
-    s3_delete_all_versions,
     s3_copy,
     s3_create_bucket,
     s3_put_cors,
@@ -180,7 +179,7 @@ function s3_get(
     end
 end
 
-s3_get(bucket, path; kwargs...) = s3_get(global_aws_config(), bucket, path; kwargs...)
+s3_get(a...; b...) = s3_get(global_aws_config(), a...; b...)
 
 """
     s3_get_file([::AbstractAWSConfig], bucket, path, filename; [version=], kwargs...)
@@ -205,9 +204,7 @@ function s3_get_file(
     end
 end
 
-function s3_get_file(bucket, path, filename; kwargs...)
-    return s3_get_file(global_aws_config(), bucket, path, filename; kwargs...)
-end
+s3_get_file(a...; b...) = s3_get_file(global_aws_config(), a...; b...)
 
 function s3_get_file(
     aws::AbstractAWSConfig,
@@ -260,9 +257,7 @@ function s3_get_meta(
     return Dict(r.headers)
 end
 
-function s3_get_meta(bucket, path; kwargs...)
-    return s3_get_meta(global_aws_config(), bucket, path; kwargs...)
-end
+s3_get_meta(a...; b...) = s3_get_meta(global_aws_config(), a...; b...)
 
 function _s3_exists_file(aws::AbstractAWSConfig, bucket, path)
     q = Dict("prefix" => path, "delimiter" => "/", "max-keys" => 1)
@@ -386,7 +381,7 @@ function s3_exists(aws::AbstractAWSConfig, bucket, path; version::AbstractS3Vers
         s3_exists_unversioned(aws, bucket, path)
     end
 end
-s3_exists(bucket, path; kwargs...) = s3_exists(global_aws_config(), bucket, path; kwargs...)
+s3_exists(a...; b...) = s3_exists(global_aws_config(), a...; b...)
 
 """
     s3_delete([::AbstractAWSConfig], bucket, path; [version], kwargs...)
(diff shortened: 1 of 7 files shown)


test/resources/awss3_jl_test.yaml
```diff
@@ -236,6 +236,38 @@ Resources:
               - s3:DeleteObjectVersion
             Resource: !Sub arn:aws:s3:::${BucketPrefix}*/*
 
+  DeleteAllVersionsTestsetRole:
```
hannahilea, line 239: @omus I am _very_ unsure if this is sufficient to add new test permissions

src/AWSS3.jl
```diff
@@ -410,6 +410,43 @@ end
...
+"""
+function s3_delete_all_versions(aws::AbstractAWSConfig, bucket, path; kwargs...)
+    # Because list_versions returns ALL keys with the given _prefix_, we need to
+    # restrict the results to ones with the _exact same_ key.
+    versions = [x["VersionId"] for x in s3_list_versions(bucket, path) if x["Key"] == path]
+    for version in versions
+        try
+            s3_delete(bucket, path; version, kwargs...)
+        catch e
+            @error "Error deleting version $(version) of $(path)\n\n$(e)"
```
ararslan, line 440: Typically for logging exceptions you do something like the following rather than interpolating the exception object into a string:
```suggestion
            @error "Error deleting version $(version) of $(path)" exception=(e, catch_backtrace())
```

src/AWSS3.jl
```diff
@@ -410,6 +410,43 @@ end
...
+# Permissions
+
+- [`s3:DeleteObject`](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html#amazons3-DeleteObject)
+- [`s3:ListBucketVersions`](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html#amazons3-ListBucketVersions)
+"""
+function s3_delete_all_versions(aws::AbstractAWSConfig, bucket, path; kwargs...)
+    # Because list_versions returns ALL keys with the given _prefix_, we need to
+    # restrict the results to ones with the _exact same_ key.
+    versions = [x["VersionId"] for x in s3_list_versions(bucket, path) if x["Key"] == path]
+    for version in versions
```
ararslan, line 436: Note that you can avoid materializing an array of version identifiers by looping over the full set of versions and skipping ones that don't meet your criteria:
```suggestion
    for object in s3_list_versions(aws, bucket, path)
        object["Key"] == path || continue
        version = object["VersionId"]
```
Also `s3_list_versions` should take the same configuration object passed to the calling function to ensure the right permissions are used.

src/AWSS3.jl
```diff
@@ -410,6 +410,43 @@ end
...
+        try
+            s3_delete(bucket, path; version, kwargs...)
```
ararslan, line 438:
```suggestion
            s3_delete(aws, bucket, path; version, kwargs...)
```
Same comment as above regarding permissions

ararslan:
Another way to implement this would be to reuse `s3_purge_versions`, which deletes all but the most recent version, then call `s3_delete` once for the remaining version. Alternatively, you could add an option to `s3_purge_versions` to include the most recent.

hannahilea:
I thought about using `s3_purge_versions` initially, but that purges versions of everything under the prefix, rather than just a specific path---and I very intentionally just want to be touching the specific path here.

ararslan:
Oh my b, I assumed you could provide a full path as the prefix to `s3_purge_versions`.

test/resources/awss3_jl_test.yaml
omus: This is perfect. If we wanted to go to the extreme we could use a more restrictive permissions set when testing just `s3_delete_all_versions` but I think this is fine

omus:
> > I thought about using `s3_purge_versions` initially, but that purges versions of everything under the prefix, rather than just a specific path---and I very intentionally just want to be touching the specific path here.
>
> Oh my b, I assumed you could provide a full path as the prefix to s3_purge_versions.

We should make an issue to make it clear on a convention for what functions deal with keys and which deal with prefixes. Right now it isn't clear and accidental usage of a prefix function when you though a key was used could be quite dangerous.

omus:
I've applied the CFN changes using admin powers. Any additional changes there will need me to update the stack

omus:
bors try

hannahilea:
> I assumed you could provide a full path as the prefix

So you can provide a full path...but then everything under it gets deleted as well. 

E.g., `s3_purge_versions("my-bucket", "foo")` will delete all [but the latest version of] "my-bucket/foo" but also "my-bucket/foo/rad/etc". This new function :just: deletes the versions of the one possible object at _exactly_ prefix "foo", and not also every prefix that _starts with_ "foo".

src/AWSS3.jl
```diff
@@ -410,6 +410,44 @@ end
...
+
+- [`DeleteObject`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html)
+- [`ListObjectVersions`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html)
+
+# Permissions
+
+- [`s3:DeleteObject`](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html#amazons3-DeleteObject)
+- [`s3:ListBucketVersions`](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html#amazons3-ListBucketVersions)
+"""
+function s3_delete_all_versions(aws::AbstractAWSConfig, bucket, path; kwargs...)
```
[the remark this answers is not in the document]
hannahilea: I've renamed it to `s3_nuke_object_versions` to prevent path v prefix confusion around the difference in behavior from `s3_purge_versions` (discussed in previous comments).

...could also switch to `s3_nuke_object`....

src/AWSS3.jl
```diff
@@ -410,6 +410,44 @@ end
...
+    # Because list_versions returns ALL keys with the given _prefix_, we need to
+    # restrict the results to ones with the _exact same_ key.
+    for object in s3_list_versions(aws, bucket, path)
+        object["Key"] == path || continue
+        version = object["VersionId"]
+        try
+            s3_delete(aws, bucket, path; version, kwargs...)
+        catch e
+            @error "Error deleting version $(version) of $(path)" exception=(e, catch_backtrace())
+        end
```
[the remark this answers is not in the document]
hannahilea: yeah, that's a great point. i think we shouldn't bother wrapping this in a try/catch at all, and just let normal exceptions occur.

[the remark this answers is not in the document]
hannahilea: ...nm, kept try/catch, but `rethrow(e)` now after printing specific error message

test/awss3.jl
```diff
@@ -354,6 +354,35 @@ function awss3_tests(base_config)
         @test s3_get(config, bucket_name, "key3") == b"data3.v3"
     end
 
+    is_aws(base_config) && @testset "Delete All Versions" begin
```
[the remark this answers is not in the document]
hannahilea: yep---all versioning tests exclude minio (s3_purge_versions, etc).

src/AWSS3.jl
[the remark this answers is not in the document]
hannahilea: (okay, there's now a naming mismatch between this function and its aws resources---i'll rename those once the updated name is deemed sufficient)

hannahilea:
bors try

hannahilea:
bors try

src/AWSS3.jl
```diff
@@ -410,6 +411,47 @@ end
 
 s3_delete(a...; b...) = s3_delete(global_aws_config(), a...; b...)
 
+"""
+    s3_nuke_object_versions([::AbstractAWSConfig], bucket, path; kwargs...)
+
+Deletes all versions of object `path` from `bucket`, forwarding all `kwargs` to the
+internal per-version [`s3_delete`](@ref) call. In the case of an error, any versions
+already deleted will be lost.
```
omus, line 419:
```suggestion
Deletes all versions of object `path` from `bucket`. All provided `kwargs` are forwarded to
[`s3_delete`](@ref). In the event an error occurs any object versions already deleted by
`s3_nuke_object_versions` will be lost.
```

src/AWSS3.jl
```diff
@@ -410,6 +411,47 @@ end
...
+
+To only delete one specific version, see [`s3_delete`](@ref); to delete all versions
+EXCEPT the latest version, see [`s3_purge_versions`](@ref); to delete all versions
+in an entire bucket, see [`AWSS3.s3_nuke_bucket`](@ref).
```
omus, line 423:
```suggestion
To only delete one specific version, use [`s3_delete`](@ref); to delete all versions
EXCEPT the latest version, use [`s3_purge_versions`](@ref); to delete all versions
in an entire bucket, use [`AWSS3.s3_nuke_bucket`](@ref).
```

hannahilea:
bors try

src/AWSS3.jl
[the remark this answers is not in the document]
hannahilea: rename complete! (resources also renamed, so may need another @omus push...)

omus:
Just updated the CFN stack

omus:
bors try

omus (approved):
image

PR #308, opened 2024-10-04
Implement multipart copy and copying a particular version

ararslan:
I've had these changes locally for months (possibly a year or more?) but hadn't committed or pushed them. I don't know if/when I'll have the bandwidth to ensure this gets over the finish line, so if someone is interested in picking this up then please feel free to do so.

Summary of changes:
- `s3_copy` now supports a `version` keyword argument that facilitates copying a specified version of an object.
- A new function `s3_multipart_copy` to mirror `s3_multipart_upload` has been added, which calls `UploadPartCopy` in the API.
- An explicit `cp(::S3Path, ::S3Path)` method has been implemented, which avoids the fallback `cp(::AbstractPath, ::AbstractPath)` method that reads the source file into memory before writing to the destination.
    - To avoid breaking the convenient but possibly unintended prior behavior of using different credentials for the source and destination paths, the fallback method is called when the source and destination credentials differ.
- `cp(::S3Path, ::S3Path)` allows the user to opt into a multipart copy, in which case multipart is used when the source is larger than the specified part size (50 MiB by default). A multipart copy is unconditionally used when the source is at least 5 GiB. This behavior mimics that of the AWS CLI. Note that this now requires an additional API call to `HeadObject` in order to retrieve the source size.

### DIFF
--- a/src/AWSS3.jl
+++ b/src/AWSS3.jl
@@ -455,7 +455,7 @@ end
 """
     s3_copy([::AbstractAWSConfig], bucket, path; acl::AbstractString="",
             to_bucket=bucket, to_path=path, metadata::AbstractDict=SSDict(),
-            parse_response::Bool=true, kwargs...)
+            parse_response::Bool=true, version=nothing, kwargs...)
 
 Copy the object at `path` in `bucket` to `to_path` in `to_bucket`.
 
@@ -464,6 +464,8 @@ Copy the object at `path` in `bucket` to `to_path` in `to_bucket`.
     See [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl).
 - `metadata::Dict=`; `x-amz-meta-` headers.
 - `parse_response::Bool=`; when `false`, return raw `AWS.Response`
+- `version=`; when not `nothing`, the specific `versionId` of the source object to copy,
+  otherwise the latest version is copied.
 - `kwargs`; additional kwargs passed through into `S3.copy_object`
 
 # API Calls
@@ -483,6 +485,7 @@ function s3_copy(
     to_bucket=bucket,
     to_path=path,
     metadata::AbstractDict=SSDict(),
+    version::AbstractS3Version=nothing,
     parse_response::Bool=true,
     kwargs...,
 )
@@ -495,10 +498,15 @@ function s3_copy(
         headers["x-amz-acl"] = acl
     end
 
+    source = "$bucket/$path"
+    if version !== nothing
+        source *= "&versionId=" * escapeuri(version)
+    end
+
     response = S3.copy_object(
         to_bucket,
         to_path,
-        "$bucket/$path",
+        source,
         Dict("headers" => headers);
         aws_config=aws,
         kwargs...,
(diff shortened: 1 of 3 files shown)


ararslan:
bors try

mattBrzezinski:
bors try

src/AWSS3.jl
```diff
@@ -1075,6 +1083,31 @@ function s3_upload_part(
...
+function s3_upload_part_copy(
+    aws::AbstractAWSConfig,
+    source,
+    upload,
+    part_number,
+    byte_range;
+    args=Dict{String,Any}(),
+    kwargs...,
+)
+    args["x-amz-copy-source-range"] = string(first(byte_range), '-', last(byte_range))
```
ararslan, line 1095:
```suggestion
    args["x-amz-copy-source-range"] = string("bytes=", first(byte_range), '-', last(byte_range))
```
https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html#API_UploadPartCopy_RequestSyntax

ararslan:
Relevant: https://github.com/JuliaCloud/AWS.jl/issues/695

src/AWSS3.jl
```diff
@@ -1145,6 +1178,70 @@ function s3_multipart_upload(
...
+    part_size = part_size_mb * 1024 * 1024
+
+    source = bucket * '/' * path
+    if version !== nothing
+        source *= "&versionId=" * version
+    end
+
+    upload = s3_begin_multipart_upload(aws, bucket, path)
+    tags = map(enumerate(0:part_size:file_size)) do (part, byte_offset)
+        byte_range = byte_offset:min(byte_offset + part_size - 1, file_size)
```
ararslan, line 1238:
```suggestion
        byte_range = byte_offset:(min(byte_offset + part_size, file_size) - 1)
```
Since it's 0-based

src/AWSS3.jl
ararslan:
```suggestion
    headers = Dict(
        "x-amz-copy-source-range" => string(
            "bytes=", first(byte_range), '-', last(byte_range)
        )
    )
    mergewith!(_merge, args, Dict("headers" => headers))
```
Otherwise it gets added as a query parameter rather than a header

src/AWSS3.jl
```diff
@@ -1145,6 +1178,70 @@ function s3_multipart_upload(
...
+"""
+    s3_multipart_copy(aws::AbstractAWSConfig, bucket, path; to_bucket=bucket, to_path=path,
+                      part_size_mb=50, version=nothing, parse_response::Bool=true,
+                      kwargs...)
+
+Copy the object at `path` in `bucket` to `to_path` in `to_bucket` using a
+[multipart copy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html).
+
+# Optional Arguments
+- `part_size_mb`: maximum size per uploaded part, in mebibytes (MiB).
```
ararslan, line 1190: I wonder if it's worth exposing an option that allows matching the part size between the source and destination. IIUC, that should make the range-based accesses faster while copying. If a file is big enough for a multipart copy, it was probably uploaded with a multipart upload, in which case the parts and their sizes can be obtained with `S3.get_object_attributes`. Lacking that permission, one can also get the part size with `S3.head_object` by passing `Dict("partNumber" => 1)` as a query parameter, and the number of parts will be in the entity tag of the source object.

